PR #37, opened 2026-07-15
fix: use analysis header on analysis-mode status screens

dinho149:
## Problem

Going to **Analysis** mode and landing on certain status/error screens showed the blue **Planning** ASCII header instead of the green **Analysis** one.

## Root cause

The wordmark helpers `analysis_title()` / `planning_title()` are correct. The bug was at the call sites: `_build_project_export_success_screen(...)` picks its title from a `mode` parameter that **defaults to `"planning"`**:

```python
title = analysis_title() if mode == "analysis" else planning_title()
```

Five calls inside the dedicated Team Analysis route (`src/scrum_agent/ui/mode_select/__init__.py`, lines ~2551–3523) omitted `mode="analysis"`, so they fell back to the blue Planning header.

## Fix

Add `mode="analysis"` to the five affected calls, matching the already-correct call at line 494:

| Screen (`subtitle=`) | Fixed |
|----------------------|-------|
| `"Board required"` (shown when entering Analysis with no tracker configured) | ✅ |
| `"Team profile exported"` (×3) | ✅ |
| `"Analysis failed"` | ✅ |

The similar calls in the Planning route's "calibrate planning with a team analysis" sub-flow are intentionally left unchanged — the blue Planning header is correct there.

## Verification

- `make lint` — clean
- `make test` — 3388 passed, 7 skipped

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "scrum-agent"
-version = "2.4.0"
+version = "2.4.1"
 description = "yeaboi.ai — a team lead's best friend. A terminal AI agent that decomposes projects into epics, user stories, tasks, and sprint plans."
 readme = "README.md"
 requires-python = ">=3.11"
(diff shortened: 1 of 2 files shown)


dinho149:
Superseded by #38. The analysis-header fix (mode="analysis") from this PR has been folded into #38, which also animates those screens (including the 'Board required' screen). Closing to keep a single comprehensive PR for the planning/analysis header work.

PR #38, opened 2026-07-15
fix: animate (shimmer) the planning and analysis page titles

dinho149:
## Problem

The **Analysis** results screen and the **Planning** session screens render their ASCII title statically, while Retro / Reporting / Standup animate a travelling "shimmer" highlight. (Standup was already correct; this PR does not touch it.)

## Root cause

The shimmer only animates when a title builder is passed a live `shimmer_tick` (a monotonic elapsed clock) each frame — `build_ascii_title` renders a solid colour when `shimmer_tick is None`. The analysis and planning builders called `analysis_title()` / `_planning_title()` with **no tick**, so the highlight never moved even though their loops already re-render every frame.

## Fix

Thread an optional `shimmer_tick: float | None = None` (default `None` → static, so **snapshots are unchanged**) through:

- **Analysis**: `_build_team_analysis_screen`, fed from its 3 result render loops.
- **Planning**: every screen builder — description input, intake question + accordion, review summary, pipeline review, chat, edit prompt, and the story / task / sprint / feature / analysis editors — each fed a per-frame `time.monotonic()` clock from its driving loop.
- **Processing screens** reuse their existing spinner `tick`, so the title keeps shimmering during LLM calls (no freeze-then-resume).

One small loop fix: choice-type intake questions now re-render on the idle timeout (previously `elif choices: continue` skipped the redraw), so the shimmer stays live while a choice question is on screen.

Left intentionally static: transient one-shot confirmation/dismiss screens (e.g. "Epic created") that render once and wait for a keypress.

## Verification

- `make lint` — clean
- `make test` — 3388 passed, 7 skipped; **19 snapshots unchanged** (confirms static default preserves existing output)
- Manual: in Analysis results and each Planning screen, the title's highlight now sweeps like Retro/Reporting.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "scrum-agent"
-version = "2.4.0"
+version = "2.4.1"
 description = "yeaboi.ai — a team lead's best friend. A terminal AI agent that decomposes projects into epics, user stories, tasks, and sprint plans."
 readme = "README.md"
 requires-python = ">=3.11"
--- a/src/scrum_agent/ui/mode_select/__init__.py
+++ b/src/scrum_agent/ui/mode_select/__init__.py
@@ -2571,6 +2571,7 @@ def select_mode(
                             height=h,
                             subtitle="Board required",
                             hint="Press any key to go back.",
+                            mode="analysis",
                         )
                     )
                     while True:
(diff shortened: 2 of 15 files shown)


dinho149:
Update: this PR now also folds in the analysis-header fix (`mode="analysis"`, previously #37, now closed) and animates the no-tracker **"Board required"** screen — its dismiss loop re-renders each frame so the green ANALYSIS title shimmers instead of showing a static PLANNING header. Also animates the analysis profile-list reveal intro.

This makes the branch self-contained: correct green ANALYSIS header **and** shimmer on every analysis entry path (with or without a Jira/AzDO tracker).

`make lint` clean; `make test` → 3388 passed, 19 snapshots unchanged.

PR #113, opened 2026-07-31
fix(ci): pin the dated haiku model id, and apply the version bump it missed

neakoh:
## The bug

`Decide & apply version bump` has failed on **every run since 2026-07-30** (last green: 2026-07-29 12:37). Evidence it's the model alias, not the credential:

- Both workflows using `--model claude-haiku-4-5` fail; all four using `claude-sonnet-5` pass — same `CLAUDE_CODE_OAUTH_TOKEN`.
- The run dies on turn 1 in ~2s with `total_cost_usd: 0` — no assistant message at all, so it never reached the model.
- Neither workflow file changed in the window that broke. #110 touched `auto-version.yml` but not the model line.
- The action prints `Running Claude Code via SDK (full output hidden for security)`, so the only visible symptom is a bare `is_error: true`.

Both haiku callers now pin `claude-haiku-4-5-20251001`.

## The missed bump

Because the bot was down, #112 merged with no bump. `main` sat at **2.41.0** — already tagged and released on 2026-07-28 — so `publish.yml` would not have shipped any of that work. Applying by hand what the bot would have: `minor` (new user-facing capability) → **2.42.0**, plus the changelog entry.

`uv.lock` re-syncs to 2.42.0 alongside `pyproject.toml`.

## Verification

`ruff check` + `ruff format --check` clean; changelog data integrity, changelog screen, update-check and version-row tests pass (77). Full suite runs in CI.

If this PR's own auto-version run goes green, the fix is confirmed end to end.


neakoh:
Update: the changelog-only experiment didn't work, and it exposed a second constraint worth recording.

`claude-code-action` refuses to run when the PR's workflow file differs from the default branch's copy:

> Workflow validation failed. The workflow file must exist and have identical content to the version on the repository's default branch.

It exits as a **skip**, so the job reports green with the model never called. Two consequences:

1. **The model pin can't be verified in this PR** — any PR that edits `auto-version.yml` is exactly the PR where the action won't run. It'll be proven or disproven on the next ordinary PR after this merges.
2. **The bot can't write the 2.42.0 changelog entry here**, so the hand-written one is restored.

Also correcting an overstatement in the description: only `auto-version` has actually run since the break. `flaky-test-hunter` (the other haiku caller) last ran 2026-07-27, before it, so it is untested rather than confirmed-failing.

PR #120, opened 2026-07-31
fix(ci): unbreak the auto-version bump, and make it report its own errors

dinho149:
## Summary

`auto-version` has been failing for **every** PR that touches `src/yeaboi/` since ~11:24 today. It looked healthy because every green run since then had *skipped* the Claude step (docs/CI-only PRs); every run that actually invoked it failed.

| Run | Branch | Claude step | Result |
|---|---|---|---|
| 17:25 | remove-local-sharing | skipped | ✅ |
| 17:15 | diag/claude-ci-error | skipped | ✅ |
| 17:06 | feature/demo-recorder | skipped | ✅ |
| 16:21 | feature/readme-banner | skipped | ✅ |
| 15:36 | remove-local-sharing | **ran** | ❌ |
| 15:46 → 17:33 (×3) | performance-coming-soon (#116) | **ran** | ❌ |

Failure signature every time: turn 1, ~2s, `total_cost_usd: 0`, `is_error: true`, **no message**. That is the same signature the file's own comment records from the 2026-07-29 alias outage — except the dated pin that fixed it last time is already in place, so the pin is not the whole story.

Meanwhile `claude-review.yml` (Opus-tier, **same** `CLAUDE_CODE_OAUTH_TOKEN`) succeeded at 17:11 and 17:19 and failed at 16:26 and 17:29 — so the token is valid and the failure is not account-wide.

## This PR

Commit 1 turns on `show_full_output: true`, keeping Haiku, so the run reproduces the failure **and prints the real SDK error** instead of a bare `is_error`. A follow-up commit applies the actual fix once that error is readable.

Keeping the diagnostic on permanently is the point: the prompt and diff here are public repo content, so there is nothing to hide, and both outages so far cost a round of blind elimination that a single error line would have prevented.

## How this gets tested without touching src/

A workflow-only PR would normally skip the Claude step and prove nothing. The guard has an explicit escape for exactly this — step 2, "A `semver:*` / `release:skip` label forces the LLM step to run" — and the workflow already triggers on `labeled`. Adding `semver:none` to this PR exercises the step with no shipping-code change.

## Test plan

- Label this PR `semver:none` → `auto-version` runs the Claude step → read the now-visible SDK error
- Apply the fix, confirm the step passes
- Re-run `auto-version` on #116 (the first PR blocked by this) and confirm it bumps + writes its changelog entry

### DIFF
--- a/.github/workflows/auto-version.yml
+++ b/.github/workflows/auto-version.yml
@@ -152,6 +152,13 @@ jobs:
         uses: anthropics/claude-code-action@v1
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+          # Show the SDK's own output. Without this the action prints only
+          # `Running Claude Code via SDK (full output hidden for security)` and a
+          # bare `is_error: true` with no message — which is why the two model
+          # outages below each took a round of blind guessing to diagnose. The
+          # prompt and diff here are public repo content, so there is nothing to
+          # hide; keep this on so the next failure reports itself.
+          show_full_output: true
           # Semver classification is mechanical (labels always win; the bump is a
           # script; the changelog paragraph is short) — routed to Haiku for cost.
           # If wrong-level bumps appear, promote to claude-sonnet-5 (one line).
--- a/.github/workflows/auto-version.yml
+++ b/.github/workflows/auto-version.yml
@@ -169,6 +169,17 @@ jobs:
           # outages below each took a round of blind guessing to diagnose. The
           # prompt and diff here are public repo content, so there is nothing to
           # hide; keep this on so the next failure reports itself.
+          #
+          # THIS FLAG CANNOT BE TESTED ON THE PR THAT ADDS IT. The action
+          # validates that the workflow file is byte-identical to the copy on the
+          # default branch, and *skips itself* when it isn't:
+          #   "Skipping action due to workflow validation: … must have identical
+          #    content to the version on the repository's default branch"
+          # The skip reports the step as SUCCESS, so a PR editing this file gets a
+          # green auto-version that never ran Claude at all. Two consequences:
+          #   1. any change here is only exercised after it lands on main;
+          #   2. a green check on a workflow-editing PR proves nothing — read the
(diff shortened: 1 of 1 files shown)


dinho149:
## Diagnosis update — this PR cannot test itself

Captured the real error by temporarily putting `show_full_output: true` on #116 (a PR that does *not* otherwise touch this workflow), then dropping the commit:

```
##[warning]Skipping action due to workflow validation: Workflow validation failed.
The workflow file must exist and have identical content to the version on the
repository's default branch.
Attempt 1 failed: ... Error is not retryable, giving up immediately
Exiting due to workflow validation skip
##[end-action ...;outcome=success;conclusion=success;duration_ms=1924]
```

**The action validates this file against `main` and skips itself when they differ — and the skip reports the step as `success`.**

Two consequences, both now documented in the file:

1. **This PR's own green `auto-version` check proves nothing.** Run `30652324713` shows `Classify & bump with Claude: success`, but the log has the validation warning — Claude never ran. Same for the `semver:none` label experiment. I initially read that green as "Haiku works fine, the outage was transient"; that was wrong, and worth flagging because it is a trap anyone touching this workflow will hit.
2. **`show_full_output` only takes effect once this lands on main.** There is no way to exercise it from a PR.

### So what was the original failure?

Still open. On #116 with this file untouched (so validation passes) the Claude step fails at turn 1, ~1.9s, `total_cost_usd: 0`, `is_error: true`, no message — three times, and once on `remove-local-sharing` too. That is the same signature this file's comment records for the 2026-07-29 alias outage, except the dated pin that fixed it is already in place.

I deliberately have **not** changed the model. The evidence that pointed at Haiku was the misread green above; with that removed there is nothing implicating the model specifically, and swapping it would be a guess that also raises cost. The sequence that actually resolves this:

1. merge this PR (guard fix + `show_full_output` + the docs)
2. re-run `auto-version` on #116
3. read the now-visible SDK error and fix the real cause

### Also in this PR

The merge-base guard fix is independent of all of the above and stands on its own — see the commit message. It is the reason #116 would otherwise have silently skipped its version bump once main moved to 2.44.0.

PR #133, opened 2026-08-02
ask for the standup transcript instead of waiting for one

dinho149:
## Summary

Standup could already read a meeting transcript, work out what the report
missed, diagnose why, and file the product-level causes as deduped GitHub
issues. What it could not do was **ask for one**. Discovery was "drop a file
into `~/.yeaboi/transcripts`, a folder you have to already know about";
`transcript_dir` was settable over MCP and nowhere else; and a team that never
learned any of this got a feature that silently did nothing, forever.

This PR is that whole loop — the review pipeline plus everything in front of it.

**Getting a transcript in.** Paste from the clipboard, `yeaboi standup-review -`
(stdin), `--transcript-text`, a bare positional path, `transcript_text` over
MCP, and a one-key reveal of the folder in Finder. Dragged paths resolve:
Terminal quotes them, iTerm2 backslash-escapes them, and both used to 404 with
no explanation. `import_text` puts the resolved date in the *filename*, so the
existing date attribution picks it up on the later sweep with no sidecar
metadata and no new schema.

**Never hunting for the folder.** Zoom, Google Meet (including the
`CloudStorage` Drive variant) and `~/Downloads` are detected and offered by
name with file counts and a newest date. Granola and Obsidian are deliberately
excluded — a proprietary cache would feed garbage to an LLM call, and a vault
of `.md` notes would mean reviewing someone's diary. The TUI can finally save
`transcript_dir` and `transcript_review_enabled`.

**Noticing the miss.** A derived set difference between "standups that ran" and
"dates a transcript was reviewed for" — no new column, no schema bump, no "last
nudged" state to migrate. The ladder is quiet during a grace period, then a
TUI-only `invite`, then a `reminder` that also enters `report.warnings` (which
broadcasts to Slack, email and exports), then `escalated`, which offers to turn
the feature **off** rather than asking again.

**Reminding.** An optional second launchd/cron job firing 30 min - 2 h after
the standup, notification-only, which posts nothing when nothing is unchecked.

### Four traps this was designed around, each verified in the source

- **Bracketed paste destroys transcripts.** `_input.py` strips every newline
  and caps at 10k chars; `_SPEAKER_RE` is line-anchored, so a pasted standup
  parses as one unattributed turn and silently narrows what the review may
  conclude. Paste therefore reads the clipboard directly.
- **The per-sweep caps took the oldest files.** Correct for the windowed
  automatic sweep, a disaster for an on-demand review with no window — the
  first run would spend its whole LLM budget on the oldest meetings on disk,
  which folder auto-detection makes likely.
- **Consent couldn't be persisted.** The popup returns
  allow-once/always/deny; the wrapper flattened it to a bool. Saving a folder
  behind an allow-once grant yields a config that works today and silently
  reviews nothing on every scheduled run after. It now refuses and says why.
- **Two OS jobs per session is a teardown problem.** Miss one path and
  disabling your standup leaves notifications firing, so `remove_schedule`
  removes *every* kind by default and the cron markers provably cannot collide.

### Review round

An independent fresh-context review found six defects, all fixed in `72ebae3`
with a regression test each — every one a "the feature quietly does nothing"
bug, which is the exact failure mode this PR exists to end. Notably: sweep
warnings were dropped whenever nothing was reviewable (so a folder that had
started being refused stayed silent); the nudge rendered only when no review
existed, hiding the whole ladder from anyone who tried this once and stopped;
a second standup on the same day could ratchet a TUI-only invite into a Slack
broadcast; and the action bar budgeted against the console instead of the panel
interior, drawing a bank of buttons off the panel at 87-88 and 91-96 columns —
selectable with the arrow keys, invisible on screen.

### Notes

- Rebased onto `main` after #131. Main took schema **v21** for artifact edits,
  so the transcript migration is renumbered to **v22**. `yesterday_context` now
  carries two distinct correction concepts that collided on one name: main's
  hand-edit log (keyword `corrections` → a `corrected` field list) and this
  branch's meeting statements (`transcript_corrections` → a `corrections`
  list). Both kept, both tested, both prompt bullets retained.
- Imports make it easy to accumulate PII in `~/.yeaboi/transcripts`, and
  **nothing in the codebase ever cleans that folder**. A "Clear reviewed
  transcripts" action is a reasonable follow-up, deliberately not in scope here.
- `file_issues` still writes to a public repo only through one explicitly
  confirmed act; nothing on a scheduled path can reach it.

## Test plan

- `make test` — **8448 passed, 47 skipped** (unit + integration + contract)
- `make lint` — clean
- End-to-end in a temp `YEABOI_HOME`: `--list-gaps` showed 6 unchecked
  standups at reminder level; piping a transcript for yesterday imported it and
  dropped the nudge to 5 at invite level (self-clearing works); an in-process
  MCP round-trip confirmed `nudge` serialises as a dict — and fixed the same
  latent bug on `latest_review`, which had been reaching agents as a `str()`
  repr because `to_jsonable` only unwraps a top-level dataclass.
- The action-bar fix was verified by reverting it and watching the new sweep
  test fail at exactly the predicted widths.
- Not exercised: the live TUI. Screens are covered by render tests, but
  `make run` → Standup › Review is worth a look before merge.


dinho149:
Thanks — five of six landed in c179392, each with a regression test. One is refuted.

**1. `read_transcript` can raise — fixed as a contract, not a catch.** The
docstring was the lie, not the code. Raising is what the sweep depends on: it
turns the exception into `Could not read X: …` in `report.warnings`, whereas
swallowing it would record a vanished file as reviewed-and-empty and conclude
the meeting contained nothing. The sweep's caller was already guarded; the one
that was not is `import_transcript`'s read-back, so `run_transcript_review` now
catches `OSError` there too — a disk that fills mid-import reaches an MCP client
as a warning, not a traceback.

**2. Recurrence counter — this one is wrong.** `bump_occurrence=False` in
`gap_issues.py` is deliberate. The counter is bumped at *draft* time:
`transcript_review.py` calls `store.upsert_gap_issue(...)` with the default
`True` for every gap of every review, which is what
`test_recurrence_bumps_occurrences` pins (three upserts → `occurrences == 3`).
Passing `True` again when filing would double-count a gap already counted when
it was drafted. "Seen 1 time(s)" on a first sighting is correct.

**3. Title-fallback dedup — fixed, and it was worse than described.** A *marked*
issue is now never matched on its title. The failure isn't only that the new gap
never files its own issue: `comment_on_gap` would post its evidence onto the
unrelated issue first, on a public repo. The fallback survives for issues filed
before the marker existed, which is what it was for.

**4. Date resolution — fixed, and it silently dropped transcripts.** `.json`
files now get the same `_MAX_CHARS` bound `read_transcript` applies, so the two
see byte-identical input and cannot disagree. Worth being explicit about the
consequence: the head-scanned date fell through to *mtime*, which for a file
downloaded today is today — so a JSON transcript covering a date inside the
window was windowed out and never reviewed, in a feature whose one job is not
missing one.

**5. Drive download storm — fixed by not reading at all.** Detection now uses
filename and mtime only. A preview label ("14 files, newest 2026-07-30") does
not need to be more accurate than an mtime, and the folders being probed are
ones the user has *not granted yet*. Also fixes a second bug in the same line:
`read_text()[:2000]` pulled each whole file into memory before slicing. There is
a test asserting detection opens zero files.

**Both minors fixed.** `standup-review -` on a tty now exits with "nothing is
piped in" instead of blocking forever, and `infer_date`'s mtime fallback logs
when it gives up — every other degrade path in that module already did.

Also in this push: the CodeQL alert, by making the VTT cue-timing matcher
anchor on its leading timestamp rather than matching a bare `-->` anywhere in
the line. `py/bad-tag-filter` is a false positive here (nothing in that module
filters HTML), but the imprecise regex was a real if minor bug — a spoken "we
moved the deploy from manual --> automated" was read as cue timing and dropped,
losing exactly the kind of sentence the review exists to check.

The other red check was `auto-version.yml`, which died mid-run against this diff
(`--max-turns 15`, part of it spent on a `git diff main..HEAD` that cannot
resolve in that checkout). Version bumped to 2.50.0 by hand with its changelog
entry, which is what the guard defers to.

PR #135, opened 2026-08-02
integrate gitlab: repo context, issues, and issue write-back

dinho149:
Closes the gap yeaboi already advertised: `detect_platform()` returns `"GitLab"` and intake Q16 offers GitLab as a code host, and then `_scan_repo_context()` bailed with `"GitLab not yet supported"` — so anyone answering GitLab got a silently degraded intake with no repo context at all.

Linear: [YEA-19](https://linear.app/yeaboi/issue/YEA-19/integrate-gitlab)

## Tools

`src/yeaboi/tools/gitlab.py` mirrors `tools/notion.py`'s shape — one `_MISSING_CONFIG_MSG` source of truth, a client factory that returns `None` instead of raising, one status-keyed error translator, `@tool` functions returning plain strings.

| Tool | Risk | Purpose |
| --- | --- | --- |
| `gitlab_read_repo` | READ | Project metadata + root file tree (nested subgroups supported) |
| `gitlab_read_readme` | READ | README off the default branch |
| `gitlab_list_issues` | READ | Backlog with labels and assignees |
| `gitlab_create_issue` | WRITE | Gated behind the `human_review` node, same as Jira/AzDO writes |

`_scan_repo_context` now runs the same two-call shape as GitHub (metadata + README) for the GitLab branch.

## Auth

`GITLAB_TOKEN` (a plain PAT — no OAuth callback, which is what makes it viable for a terminal app), plus `GITLAB_URL` for self-hosted instances, normalised on read like `AZURE_DEVOPS_ORG_URL`. `read_api` covers the reads; `api` is needed only for issue creation.

**These tools deliberately do not degrade to anonymous access the way the GitHub ones do.** gitlab.com *will* serve anonymous reads of public projects (verified: `GET /api/v4/projects/gitlab-org%2Fgitlab-foss` → 200 with no token), but private projects and most self-hosted instances reject them. Degrading would work for some repos and 401 for others with nothing to tell the user which happened, so a missing token is reported as "not configured" instead.

The credential is wired through every surface that handles secrets: a **Setup → Credentials → GitLab** section (token + instance URL), a `TOKEN_HELP` entry naming the scopes and the creation URL, and `GITLAB_TOKEN` plus a `glpat-` token-shape pattern in the redaction registry.

## Surface parity

No `CAPABILITIES` row — verified, not assumed. That registry's keys are *modes*; `mcp/tools_*.py` and `claude-plugin/yeaboi/skills/` are per-mode too, and no integration (GitHub, Jira, Notion, Confluence, AzDO) has a row. Agent tools reach the TUI, CLI and MCP at once because all three run the same ReAct loop. Enforcement for this layer is `test_tools_registry.py` (AST discovery vs `get_tools()`) and `test_risk.py` (two-way vs `TOOL_RISK`), both of which GitLab now satisfies.

## Tests

`tests/unit/tools/test_tools_gitlab.py` — 62 tests: happy path and error path per tool, the missing-credential path for all four, URL parsing (nested subgroups, `.git`, `/-/tree/main`), the 403-vs-404 distinction in the README candidate loop, the `per_page` clamp, and token masking.

`TestScanRepoContext` in `tests/unit/nodes/test_analyzer.py` gains GitLab coverage — its `test_gitlab_platform_returns_none` asserted the old "not supported" behaviour and is replaced by tests for the new one, including the no-token degraded path.

Local results: 467 integration passed, 190 golden passed, `ruff check` and `ruff format --check` clean, `make site-check` clean, and every unit test in the touched areas green.

Two sets of pre-existing failures in this sandbox are **not** from this branch — both reproduce identically on clean `origin/main`:
- 11 contract tests (`test_confluence_contract.py`, `test_github_contract.py`) fail where the VCR shim can't reach the network.
- `tests/unit/test_setup_wizard.py` leaks `JIRA_PROJECT_KEY` into `os.environ`, after which `test_standup_engine.py` makes a **real Jira API call** during the unit suite. Run in its own process, that file is 113 passed in 15s; run after the leak it stalls for minutes per test. Harmless in CI (fails fast), worth its own ticket.

## Out of scope

Merge requests, pipelines, milestones/epics, a file-by-path reader (`github_read_file` has no GitLab twin yet), batch board sync, and the first-run setup wizard's version-control step (still GitHub-only — GitLab tokens go in Setup → Credentials, which is also the only place that takes the self-hosted instance URL).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EX4tk842kkbsMki9V8agQx)_


dinho149:
⚠️ **Heads-up: #136 implements the same GitLab integration.** Both PRs close YEA-19 and cannot both merge.

This one was opened first (21:03 vs 21:27). Two sessions picked GitLab independently for the same reason — the `_scan_repo_context` bail-out — and landed on nearly identical designs, including the same no-anonymous-fallback call and the same "no `CAPABILITIES` row" conclusion.

Where they differ: this PR is currently `mergeable_state: dirty` (conflicts with main after #134 landed), whereas #136 is green on all 13 checks. This PR does include something #136 lacks — the **Setup → Credentials → GitLab section and `TOKEN_HELP` entry**, so the token is settable from the TUI instead of only `.env`.

Flagging rather than acting: whichever you keep, the other should be closed, and this PR's Setup → Credentials work is worth carrying over if #136 is the one that lands.

---

PR #136, opened 2026-08-02
integrate GitLab: 3 read tools + 1 write tool, wired into the intake repo scan

dinho149:
Closes YEA-19.

## Why GitLab

yeaboi already told users it supported GitLab. `detect_platform()` returns `"GitLab"`, and intake Q16 offers it as a code-host option — and then `_scan_repo_context()` bailed with `"GitLab not yet supported"`. Anyone who answered "GitLab" got a silently degraded intake with no repository context.

Every other candidate on the shortlist (Linear, Sentry, Bitbucket, Asana) would add a *new* capability. This one fixes a *broken* one, at the lowest cost — a mature SDK and plain token auth.

## What's in it

| Tool | Risk | Purpose |
|---|---|---|
| `gitlab_read_repo` | READ | Project metadata, directory tree, key tech-stack files, language breakdown |
| `gitlab_read_readme` | READ | README (+ CONTRIBUTING.md) for project analysis |
| `gitlab_list_issues` | READ | Open issues feeding planning |
| `gitlab_create_issue` | WRITE | Ticket write-back, gated behind `human_review` |

`tools/gitlab.py` mirrors the shape of `tools/github.py` and `tools/notion.py`: module docstring covering the SDK choice and auth model, a single `_MISSING_CONFIG_MSG`, `@tool`-decorated functions, and the explicit "only call after the user confirms" note on the write tool.

Also wired: the GitLab branch of `_scan_repo_context` now scans like GitHub and Azure DevOps do.

## Env vars

- `GITLAB_TOKEN` — personal access token (`read_api` for reads, `api` to create issues)
- `GITLAB_URL` — optional, defaults to `https://gitlab.com`; set for self-hosted

Both documented in `.env.example`.

## Three GitLab-specific details worth reviewing

1. **Nested subgroups.** A project can live arbitrarily deep (`group/sub/team/project`), so the tools take a `project_url`, not an owner/repo slug, and `_parse_project` cuts at GitLab's `/-/` separator rather than keeping the first two path segments.
2. **`opened`, not `open`.** GitLab returns an *unfiltered* list for an unrecognised issue state rather than erroring, so passing GitHub's `open` would silently include closed issues. The state is validated instead of forwarded.
3. **A token is required.** Unlike the GitHub tools, these do not fall back to anonymous access — GitLab's unauthenticated API is heavily restricted, and an anonymous 404 is indistinguishable from "this project is private", which would surface to the user as a confusing "not found".

## Security

`GITLAB_TOKEN` joins `SECRET_ENV_KEYS`, plus a `gl<kind>-` token pattern so a pasted `glpat-`/`glrt-` token is scrubbed from logs even when it never came from our env. Covered by masking tests.

## Tests

70 added — `tests/unit/tools/test_tools_gitlab.py` (67: happy path and error case per tool, missing-credential path, URL parsing including subgroups and self-hosted, token masking, registration, risk classification) plus 3 in `test_analyzer.py` covering the newly-wired branch.

Four existing registry assertions were updated rather than worked around: the total tool count moves 37 → 41, and `gitlab_create_issue` joins the expected WRITE set in `test_risk.py`. One obsolete test — `test_gitlab_platform_returns_none`, which pinned the exact bug being fixed — is replaced by three tests of the new behaviour.

## Notes for the reviewer

- **No `CAPABILITIES` row.** `CAPABILITIES` keys are modes, and its checks are two-way set equality against `_MODE_CARDS`, the MCP tool list, and plugin skills. Every existing tool integration (GitHub, Jira, Notion, Confluence, AzDO) has no row; adding one would break the suite. GitLab follows the same contract they do — `test_tools_registry.py` (AST discovery vs `get_tools()`) and `test_risk.py`. Parity holds because agent tools reach the TUI, CLI, and MCP at once through the shared ReAct loop.
- **Deliberately out of scope**, listed so they're visible rather than silently missing: no merge-request/pipeline reading (standup's activity feed stays GitHub-only), no batch board sync like Jira/AzDO have, and no Setup-wizard card — the token goes in `.env` like any other credential. A GitLab section for `docs/docs/integrations-exports.html` is a sensible follow-up.

---
_Generated by [Claude Code](https://claude.ai/code/session_019caHoVTBSbUMBuPfQ5PCbb)_

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,17 @@ LANGSMITH_PROJECT=yeaboi
 # No token: works for public repos at 60 req/hour (unauthenticated)
 GITHUB_TOKEN=your-github-personal-access-token
 
+# GitLab (optional — enables project context tools and issue creation)
+# Create at: GitLab → User Settings → Access Tokens
+# Scopes: read_api (read tools) or api (also create issues)
+# Unlike GitHub there is no unauthenticated fallback — without a token the
+# GitLab tools return "GitLab is not configured" rather than trying anonymously.
+# Real tokens are issued with a "glpat-" prefix; the placeholder below omits it
+# so secret scanners don't flag this example file.
+GITLAB_TOKEN=your-gitlab-personal-access-token
+# Optional: point at a self-hosted GitLab. Defaults to https://gitlab.com.
+GITLAB_URL=https://gitlab.com
+
 # Azure DevOps (optional — enables repo context tools and board sync; required for private projects)
 # PAT permissions needed: Code=Read, Work Items=Read+Write, Project=Read
 AZURE_DEVOPS_TOKEN=your-azure-devops-personal-access-token
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "yeaboi"
-version = "2.51.0"
+version = "2.52.0"
 description = "yeaboi — an AI Scrum Master for your terminal: sprint planning, standups, retros, planning poker, 1:1s and delivery reports."
 readme = "README.md"
 requires-python = ">=3.11"
@@ -46,6 +46,10 @@ dependencies = [
     # Notion doc tools (tools/notion.py) — official SDK, mirrors how Confluence
     # uses atlassian-python-api. Notion has its own integration token (NOTION_TOKEN).
     "notion-client",
+    # GitLab repo/issue tools (tools/gitlab.py) — the maintained official SDK,
+    # mirroring how GitHub uses PyGithub. Auth is a personal access token
+    # (GITLAB_TOKEN); GITLAB_URL points it at a self-hosted instance.
+    "python-gitlab",
     # SQLite-backed checkpoint store for Phase 8 session persistence.
     # SqliteSaver is used as a standalone store; the graph itself stays
     # stateless for now (Phase 8B wires it as a full LangGraph checkpointer).
--- a/src/yeaboi/agent/nodes.py
+++ b/src/yeaboi/agent/nodes.py
@@ -2437,10 +2437,10 @@ def _fetch_notion_context(
 def _scan_repo_context(questionnaire: QuestionnaireState) -> tuple[str | None, dict]:
     """Scan the repo referenced in Q17 and return a combined context string + status.
 
-    Calls GitHub or AzDO read tools directly as Python functions — no LLM
-    ReAct loop needed, they are plain functions that return strings.
+    Calls GitHub, GitLab, or AzDO read tools directly as Python functions — no
+    LLM ReAct loop needed, they are plain functions that return strings.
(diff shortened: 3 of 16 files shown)


dinho149:
Both findings were real — fixed in 973cb42.

**Changelog copy.** Confirmed and corrected. The entry was generated by the auto-version bot and I absorbed it into this branch without reading it closely enough, so thanks for catching it. While fixing the anonymous-access sentence I found a second error in the same entry: a highlight claimed you can create GitLab issues "the same way GitHub and Jira do", but the GitHub tools are read-only — `risk.py` registers no GitHub write tool at all. Both are corrected, and the token highlight now says the token is required with no anonymous fallback.

**`max_issues` above 100.** Also real. `per_page` was clamped to GitLab's server-side max while the loop and the "increase max_issues" hint still compared against the caller's `max_issues`, so `max_issues=250` returned 100 issues with `(100 issues shown)` — indistinguishable from "that's the whole list". The cap is now an explicit `_MAX_PER_PAGE` constant and the output says when it applied: `GitLab returns at most 100 per request, so 250 was capped`. Covered by a new test.

`make lint` clean, 84 tests pass locally in the touched files.

---

dinho149:
⚠️ **Duplicate of #135 — these two PRs implement the same integration and cannot both merge.**

I only discovered this after CI went green here: #135 was opened at 21:03, this one at 21:27, and both close YEA-19. Two sessions picked GitLab independently, for the same reason (the `_scan_repo_context` bail-out), and converged on nearly identical designs — same four tools, same `tools/notion.py` shape, same no-anonymous-fallback decision, same "no `CAPABILITIES` row" conclusion.

Honest comparison so someone can pick quickly:

| | #135 | #136 (this) |
|---|---|---|
| CI | `mergeable_state: dirty` — conflicts with main | **13/13 green** |
| Files / tests | 23 files, 62 tests | 17 files, 71 tests |
| Setup UI | **Adds Setup → Credentials → GitLab + `TOKEN_HELP`** | Not included — token via `.env` only |
| Issue listing | `per_page` clamp | clamp **plus** an explicit message when the 100 cap applies |
| Changelog | — | corrected (the auto-generated entry claimed anonymous access works, which is false) |

**#135 has one thing this PR does not**: the Setup → Credentials section and `TOKEN_HELP` entry, so the token is settable from the TUI rather than only `.env`. That is a genuinely better credential surface. This PR's advantages are that it is currently green and mergeable, and has the extra test coverage plus the changelog correction.

My suggestion, but it's your call: merge this one to get GitLab in green, then cherry-pick #135's Setup → Credentials section as a small follow-up. Closing whichever you don't want is a one-click decision — I'd rather flag the collision than silently race it.

---

dinho149:
Closing to reset the cowork backlog while the fleet is throttled back to a single lane. The branch `claude/ecstatic-clarke-wp3bqn` is **kept** — reopen this PR to fold the work into a later release batch.

Issue #153, opened 2026-08-05
[bug][tui-ux] provider picker crops ollama row on 24-row terminals


dinho149:
The first screen of the setup wizard — where a new user picks an AI provider — cuts off its bottom rows on a standard-sized terminal window. The parts that disappear are the free, no-API-key option and the setup progress indicator. New users on default terminal sizes may never see the one option that would get them started without a paid account.

**Effort** M

---

**What**

`_build_select_screen` (`src/yeaboi/ui/provider_select/screens/_screens.py:227`), the first-run LLM-provider picker, does not window its content when there isn't enough vertical space. At `width=80, height=24` — the builder's own real default height, since `h` comes from `console.size` at `src/yeaboi/ui/provider_select/__init__.py:207`, i.e. an ordinary 24-row terminal — rendering the screen crops away both the Ollama row (the only no-API-key, free option) and the entire progress-chip footer entirely. `_build_screen_frame` centres a fixed `body_height` of 14 into a `middle_h` of 9 with no scroll/windowing, unlike `_build_model_select_screen`, which does window its content to fit.

**Why now**

This is the very first screen a new user sees during setup. On a standard 24-row terminal (a common default, e.g. many default macOS Terminal profiles, tmux panes, CI log viewers), the option most likely to get someone unblocked without an API key — Ollama — is invisible, and so is the progress indicator that would otherwise orient them in the wizard. This is a first-impression bug for exactly the audience most likely to hit it (a new/free user on a small terminal).

**Evidence**

Found while adding render-test coverage for this screen (Linear YEA-32, PR tests the screen at `height=24` and confirms `"LLM Provider"` — the active progress chip — and the Ollama row are absent from the render at that height, versus the frame rendering the same content correctly at `height=40`). `.claude/skills/tui-standards/SKILL.md` calls out `calc_viewport`/scrollbar windowing as the standard pattern for exactly this case; `_build_model_select_screen` in the same package already does this correctly and can serve as the reference implementation.

**Effort**

M — likely needs `_build_screen_frame` (or the select-screen caller) to window/scroll the provider list the same way `_build_model_select_screen` does, using the existing `calc_viewport`/`build_scrollbar` shared primitives per tui-standards rather than a bespoke fix; touches `_screens.py` and probably `_screens_vc.py` (same underlying frame builder, same risk), plus render tests at multiple heights to pin the fix.

Approve by adding the `claude-implement` label. Reject by closing this issue.

---

[...]

PR #154, opened 2026-08-05
add render tests for provider-select and vc-select screens

dinho149:
## Summary

This week's `tui-ux` cowork sweep auto-lane pick (Linear [YEA-32](https://linear.app/yeaboi/issue/YEA-32/add-render-tests-for-provider-select-and-vc-select-screens)).

`_build_select_screen` (`src/yeaboi/ui/provider_select/screens/_screens.py:227`) and `_build_vc_select_screen` (`src/yeaboi/ui/provider_select/screens/_screens_vc.py:54`) are the LLM-provider-pick and version-control-provider-pick screens — the very first screens a new user sees during setup — and had zero render tests anywhere in `tests/unit`. This PR is a pure test addition; no production code changes.

- Adds `tests/unit/test_provider_select_pickers.py` (two classes, one per screen): returned `Panel`, subtitle row (exact-row matched, isolated from the progress chip and from other cards' taglines), every option row renders, selected-card tagline shown/hidden correctly, the `visible` reveal/retract transition window (including the empty-visible and selection-off-screen frames), `selected_style`/`fade_style`/`fade_indices` overrides on both the unselected and selected fade paths (the latter is the shape every real caller — the confirm pulse, the VC fade-in — actually uses), `shimmer_tick` animation, the `step` argument driving the active chip, and the real short-terminal invariant (progress chip and last provider row surviving a 24-row frame).
- Two tests (`TestSelectScreen::test_progress_bar_survives_a_short_terminal` and `test_last_provider_row_survives_a_short_terminal`) are `xfail(strict=True)`: independent code review of this PR found the LLM picker's progress footer *and* its last provider row are genuinely cropped off at short terminal heights (`_build_screen_frame` centres a fixed body height with no windowing). That's a real UX defect, filed separately as tui-ux proposal [#153](https://github.com/dinho149/yeaboi.ai/issues/153) / [YEA-37](https://linear.app/yeaboi/issue/YEA-37/stop-llm-provider-picker-from-cropping-ollama-row-and-progress-footer) rather than fixed here, since layout changes always propose per house-rules. Both xfails are pinned with a guard against unrelated changes (list-length drift) making them a confusing XPASS instead of a meaningful one, and will flip to hard failures the moment #153 is actually fixed.

## Test plan

- `make lint` — clean.
- `make test` — 9182 passed, 47 skipped, 2 xfailed (the two `#153` pins), 3 failed. All three failures are pre-existing and unrelated to this change (verified identical with this PR's file stashed): two `test_cowork_setup.py::TestGhWrites` tests skip/fail without `gh` on `PATH`, and `test_standup_transcripts.py::TestScanBounds::test_unreadable_directory_is_skipped_not_fatal` fails because the sandbox runs as root (a `chmod 000` directory is still readable).
- Diff went through three rounds of independent review (`code-reviewer` plus automated `claude-review.yml` passes); all should-fix and real coverage-gap findings are resolved, including anchoring style-override assertions to the specific row they claim to prove, deriving test indices from list length instead of hardcoding, pinning `no_color=False` so the SGR assertions aren't environment-dependent, and adding coverage for the fade-on-selected-row branch that production code actually exercises (confirmed via mutation testing that this — not an arm-order swap a reviewer also flagged, which is provably a no-op at every real call site — is the regression the test needed to catch).

## DoD exemptions

- Item 5 (surface parity) — no new capability, no `CAPABILITIES`/`FeatureTip` change needed.
- Item 7 (web bundles) — nothing under `frontend/`.
- Item 6 (observability) — `_build_*_screen` builders run every frame, so per tui-standards they carry no logging; this change adds only tests.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

---
_Generated by [Claude Code](https://claude.ai/code/session_012uB3gvLHc1H9ajVxzDSt5N)_

### DIFF
--- a/tests/unit/test_provider_select_pickers.py
+++ b/tests/unit/test_provider_select_pickers.py
@@ -67,6 +67,11 @@ def _render_ansi(panel: Panel, *, width: int = 80, height: int = _TALL) -> str:
         color_system="truecolor",
         legacy_windows=False,
         highlight=False,
+        # Rich reads NO_COLOR from the environment even with force_terminal +
+        # color_system set, and strips every SGR when it is present. Pinning it
+        # keeps the escape-matching assertions below from failing on the machine
+        # of anyone who sets NO_COLOR=1 while CI stays green.
+        no_color=False,
     )
     console.print(panel)
     return console.file.getvalue()
@@ -82,6 +87,22 @@ def _vc(selected: int = 0, **kwargs) -> Panel:
     """Build the version-control picker at a frame height that fits every row."""
     kwargs.setdefault("height", _TALL)
     return _build_vc_select_screen(selected, **kwargs)
+
+
+def _card(provider_val: str) -> dict:
+    """The provider card for *provider_val* — by identity, never by list index."""
+    return next(card for card in _PROVIDER_CARDS if card["provider_val"] == provider_val)
+
+
+def _tagline(card: dict) -> str:
+    """A card's tagline, read the way the screen reads it.
+
+    ``_screens.py`` uses ``.get("tagline", "")`` and falls through to the plain
+    separator when there is none, so the tests must tolerate a card without one
+    instead of raising ``KeyError`` and taking the suite down with an error
+    rather than a failure.
+    """
+    return card.get("tagline", "")
 
 
 def _art_top(name: str) -> str:
@@ -106,16 +127,23 @@ def _active_chip(out: str) -> str | None:
     return match.group(1) if match else None
 
 
-def _has_subtitle(out: str, text: str) -> bool:
-    """True when *text* appears on the frame's subtitle row, not just anywhere.
+def _text_rows(out: str) -> set[str]:
+    """The frame's content rows, stripped of the page panel's border and padding.
 
-    The VC picker's subtitle is "Version Control" and its active progress chip is
-    labelled "Version Control" too (``_STEPS[3]``), so a plain ``in`` check passes
-    with the subtitle row deleted — it only re-proves the chip. The chip row is
-    the one drawn with the ▟/▛ parallelogram caps, so excluding those lines
-    isolates the subtitle.
+    Whole rows rather than a substring scan over the whole capture, because both
+    kinds of substring collision are live here:
+
+    - across rows: the VC picker's subtitle is "Version Control" and so is its
+      active progress chip (``_STEPS[3]``), so ``"Version Control" in out`` holds
+      with the subtitle row deleted — it only re-proves the chip;
+    - within a row: Anthropic's tagline ("Recommended cloud · API key required")
+      ends with OpenAI's ("Cloud · API key required") bar the capital, so the day
+      someone normalises that copy a substring check starts claiming OpenAI's
+      tagline is on screen when it is not.
+
+    An exact row match is immune to both.
     """
-    return any(text in line and "▟" not in line and "▛" not in line for line in out.splitlines())
+    return {line.strip().strip("│").strip() for line in out.splitlines()}
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dinho149:
Thanks for the review — findings 1-5 are addressed in 176e3a2:

1. The xfail now pins both symptoms named in #153: added a second strict xfail asserting the Ollama row is also cropped at `_SHORT`, not just the footer.
2. Tagline/subtitle checks now compare whole rendered rows rather than substrings, closing both the cross-row collision (VC subtitle vs. chip) and the in-row one you flagged (Anthropic's tagline containing OpenAI's up to capitalization).
3. Tagline reads now mirror `_screens.py`'s `.get("tagline", "")` fallback instead of indexing directly.
4. Both xfail `reason=` strings now name `#153` literally.
5. `no_color=False` is pinned on the render console — confirmed it's load-bearing (the SGR assertions do fail under `NO_COLOR=1` without it, pass with it).

Left the `width` variation (narrow-terminal compact-font fallback) out — that's new coverage for a different scenario, not a fix to what's here.

`make lint` and `make test` are green (9179 passed, 2 xfailed — both new and expected, 3 pre-existing unrelated failures verified identical without this diff).

---

dinho149:
Both notes addressed in bb20fd5:

1. Style probes now anchor to the specific row via a new `_art_row` helper rather than a bare substring-anywhere scan, plus a row-count check for the fade case (the shimmering selection styles per-character, so it can't be substring-anchored the same way — verified it matches zero contiguous rows, which is what makes the count check necessary rather than redundant). Mutation-checked against the three branch bugs this was meant to catch (off-by-one, paint-every-row, ignore-selected) — all three now fail where they previously passed.
2. Hardcoded indices (`selected=4`, `_VC_OPTIONS[1]`) are now derived from `len(...) - 1` so the tests keep meaning what they claim if either list changes size.

`make lint` / `make test` green (9179 passed, 2 xfailed as expected, same 3 pre-existing unrelated failures as every prior run on this branch).

---

dinho149:
Pushed in 79c6c05, plus updated the PR description (it said "one" xfail, there are now two).

1. Added the fade-on-selected-row test to both classes. One correction to the review's mechanism, verified by mutation: swapping the `if i == selected / elif i in fading` arms is provably a no-op at every real call site, since `selected_style` is always empty exactly when the selected index is in `fade_indices`, and vice versa — so either arm order reaches the same result. The regression that *is* real, and that the new tests actually catch, is conditioning the fade on the row being unselected (e.g. `elif i in fading and fade_style and i != selected`) — a plausible "the selection already shimmers" tidy-up that would silently kill the confirm pulse and the VC fade-in.
2. Both strict xfails now carry a guard, and — since an assertion failing *inside* an xfail is just another expected failure and never surfaces its message — a companion normal test that fails loudly by name if the list-length assumption drifts, so the guard is actually visible rather than swallowed.
3. The last hardcoded index (`faded = 2`) is now derived from list length, with an explicit assert that it's non-selected.

`make lint` / `make test` green (9182 passed, 2 xfailed as expected, same 3 pre-existing unrelated failures as every prior round).

---

dinho149:
Agreed the mechanism is right and the naming/threshold split is a fair maintainability nit — `test_the_height_xfails_still_describe_a_layout_bug` bundling the VC guard (which duplicates the check already inside `TestVcSelectScreen::test_progress_bar_survives_a_short_terminal`) and using `>=5` where the footer-only xfail only needs `>=4` means a 4-card drop would red-build with a message pointing at the wrong threshold.

Leaving it as-is for this PR: it's advisory, non-blocking, and purely about which assertion a future failure message points to — this branch has already been through four review rounds tightening real coverage gaps (the fade-on-selected-row branch, row-anchored style probes, `no_color` pinning, list-length-derived indices), and this is a smaller-than-those cleanup on a guard test whose only job is to fail loudly if someone changes a list length years from now. Happy to pick it up if it comes back with another finding attached, but not spinning up a fifth round for this alone.

No other findings this round — thanks for confirming the security/conventions checks.

---

dinho149:
Closing to reset the cowork backlog while the fleet is throttled back to a single lane. The branch `cowork/tui-ux-provider-select-render-tests` is **kept** — reopen this PR to fold the work into a later release batch.

PR #215, opened 2026-08-10
begin the go migration: yeaboi-core sidecar serving agentwatch and the standup core

dinho149:
## Summary

Begins the long-term strangler-fig migration of yeaboi's deterministic engines to Go, as eight phase commits:

1. **Startup**: defer `cli.py` heavy imports + lazy-load the `agent` package (`--version` 0.65s → 0.04s, pinned by `test_cli_startup.py`).
2. **Agentwatch scan**: 15–30s cold scan → ~1.1s (regex substring guards, batched SQLite transactions, process-pool parsing).
3. **The seam**: `contracts/v1/rpc.md` (ndjson JSON-RPC), the pure-Go `yeaboi-core` binary, `src/yeaboi/gocore/` client, and agentwatch usage dispatch with silent Python fallback.
4. **Agentwatch in Go** with a byte-level Python↔Go parity suite (`tests/parity/`, `parity` CI job).
5. **The whole agentwatch family** (usage, standup, security) served from the sidecar; ordered-JSON + Python `repr`/`json.dumps` emitters (contract rules 8–9).
6. **Distribution**: `packaging/yeaboi-core/` platform wheels (5 targets, wheels-only), `yeaboi[core]` extra, `YEABOI_GO` three-state auto-detect, `publish-core.yml` on the `core-v*` tag line.
7. **The standup deterministic core** as one `standup.aggregate` RPC — identity closure → filters → coverage → grouping → insights → habits/relatedness → confidence → member skeletons, with an idempotent two-pass protocol for the practice-adjudication LLM interleave. Result key order is contractual (it feeds the LLM prompt bytes). Binary 0.2.0, contract stays v1.
8. **Ship-review fixes**: the schema guard now reads `schema_info` (the table Python actually records versions in — `PRAGMA user_version` was never set, so the guard could not fire), spawn-context + submission-order parse pool, `refresh()` never-raises under lock contention, exact-implication bearer guard in both languages, table-driven `pysem` tests generated from CPython outputs.

Python remains the reference implementation; `make parity` and the CLAUDE.md dual-maintenance section keep the twins in lockstep. The TUI is untouched. The privacy invariant holds: Go logs nothing, and no transcript/activity content reaches errors or warnings.

**Accepted findings from the independent review** (not fixed, deliberate): a persistently-crashing sidecar binary re-spawns once per engine run before falling back (correctness unaffected — every call still silently falls back to Python; a crash-count latch is future work), and the agentwatch risky-bash patterns keep RE2's ASCII `\b` (documented in `patterns.go`; every boundary sits against ASCII shell syntax).

**Manual steps this PR does not do** (needed before/after merge):
- Add a PyPI trusted-publisher entry for `yeaboi-core` pointing at `publish-core.yml`.
- Add the `go` and `parity` CI checks to the main-branch ruleset.

## Test plan

- `make test` — 10,926 passed / 47 skipped
- `make lint`, `make security` — clean
- `go build/vet/test ./...`, `gofmt` — clean (incl. new `pysem` unit tests)
- `make parity` — 22/22 (agentwatch 9 + standup 13), byte-level incl. key order
- E2E: a dry-run standup with the sidecar discovered logs "standup.aggregate served by the sidecar" and its report is byte-identical to a `YEABOI_GO=0` run; wheel smoke (build → venv install → `binary_path()` → `core.hello`) runs in the `go` CI job

DoD notes: web bundles untouched (no `frontend/` changes); no new user-facing capability rows (the sidecar is a transparent backend, `TIP_EXEMPT` not applicable); Notion/Slack fire on merge via `pr-merged-close-loop`.

Closes YEA-73


dinho149:
Thanks — the should-fix was right and is fixed in c0e76c5, along with two of the four nits.

**Fixed**

- **`discovery.py` CWD hijack (should-fix).** Confirmed against the installed CPython: `shutil.which` inserts `os.curdir` at position 0 of the search path on win32. Since discovery is automatic and silent, that is an unprompted spawn of an attacker-plantable file. A cwd hit is now rejected on *every* platform (a POSIX `PATH` containing `.` is the same hazard), with `YEABOI_CORE_BIN` as the explicit opt-in for anyone who really means the local binary. Four tests cover it: cwd-relative hit refused, absolute-path cwd hit refused, env var still reaches a cwd binary, PATH elsewhere still used.
- **Leaked child on handshake failure.** `client.close()` now runs in the `except` branch.
- **`.gitignore`** is `/bin/`, anchored to the `make go-build` output.

**Not fixing, with reasons**

- **`_parse_worker` discards malformed-line counts on the error path.** Two reasons to leave it. First, the Go twin discards it identically (`collector.go`, the `res.errClass != ""` branch skips the `stats.MalformedLines +=`), so Python and Go currently agree — changing one side alone would be a real parity divergence that the corpus would not necessarily catch. Second, discarding looks like the more correct behaviour: a file that errored gets no cursor and is fully reparsed next run, so counting its malformed lines now would double-count them then. The file itself is already named in `stats.warnings`, so nothing is silently lost.
- **`_DEFAULT_TIMEOUT = 600.0`.** Agreed it is loose, but per-method timeouts are a contract change (they belong in `contracts/v1/rpc.md` next to each method, not as a client-side constant), and the failure it would bound requires the reader-thread EOF drain to fail first — which is itself the bug to fix if it ever happens. Better as a deliberate change than as a feedback-pass tweak. Filed as follow-up work rather than dropped.

Full gate green on the fixes: `make test` 10,930 passed / 47 skipped, `make lint` clean, `make parity` 22/22.

PR #224, opened 2026-08-11
move the doc-quality scoring core behind the go sidecar

dinho149:
## Summary

Wave 6 of the Go migration: the team-analysis **documentation-quality scoring core** is now served by the yeaboi-core sidecar as one additive RPC method, `analysis.score_docs` (binary 0.4.0, contract stays v1 — an old binary answers -32601 and Python computes silently). The docs path has no LLM anywhere (`generate_doc_quality_insights` is a deterministic shim), so unlike Wave 5's two methods, a single method serves the whole tail: per-page clarity/usefulness/disclosure scoring, aggregation into the `DocQualitySignal`, findings, the action plan, and the coaching insights.

- **Python seam** (`analysis/aggregate.py`): `build_score_docs_inputs` / reference `score_docs` / `go_score_docs` / a lossless `DocQualitySignal` wire pair. The read workers stop scoring and cache-writing (`_gated_read` reads only); the score cache is written in one post-seam batch (`_write_doc_cache`, `zip(strict=True)` against `scoreable_doc_pages`, and the dispatch rejects an asset-count mismatch as malformed before it can mis-key a cache row). `collect_doc_pages`' readable filter widened to asset-or-body (equivalent set — workers only ever attached an asset when the body was non-blank). The insights coverage gate moved from `tools/team_learning.py` into `run_doc_quality` (same predicate on the same object, blob key position preserved). Dead `aggregate_doc_quality` and `_collect_samples` deleted; their tests retargeted at the live `_aggregate_doc_assets`.
- **Go twin** (`go/internal/analysis/doc_quality.go`): hand-rolled scanners where RE2 diverges from Python — unicode-space sentence splitting (Python re `\s` == `str.isspace()` == `pysem.IsSpace`, enumerated over all of Unicode), per-branch disclosure matching, per-word vocabulary scans that survive the multi-word `next step` alternative, unicode digits for list markers, `İ`-aware lowering, banker's rounding and Python float repr throughout — and explicit float64 conversions around the Flesch expression so arm64 FMA fusion cannot flip a rounding tie the amd64 parity job would never observe. The package still imports no `log`, now pinned by a test since page bodies cross the wire.
- **Tests**: `tests/unit/test_doc_scoring.py` (72 exact-value baseline tests written before the port), seam tests in `test_analysis_aggregate.py`, 38 Go tests pinning Python-generated bytes, and `tests/parity/test_docs_parity.py` (NBSP, Turkish İ, Arabic-Indic list, owner-across-blank-lines, cached-asset passthrough, 80-code-point truncation; self-guards run unskipped in the ordinary suite). Whole-seam byte parity: **36/36**.

Three deliberate trade-offs from the hoist, disclosed rather than hidden: an interrupted run no longer checkpoints partial scores mid-read; reporting's `collect_doc_pages` call no longer warms the score cache as a side effect; and a raise while scoring now fails the whole Documentation component (best-effort empty signal) instead of one page.

## Rebase note (2026-08-17, go-migration campaign)

This PR predates the go-migration workstream; the campaign's bootstrap run adopted it: labelled `cowork` + `workstream:go-migration` + `type:chore` (arming the parity hold — `PARITY_GATED_PRS` gates this PR by number since the branch predates the `cowork/migration-w` prefix), rebased on `origin/main` (conflicts: `uv.lock` regenerated, not hand-merged; CLAUDE.md's dual-maintenance entry re-expressed in main's new table form with the same enumerated `doc_quality.py` pieces), bumped the version 3.17.0 → 3.18.0 (wave PRs bump **minor**; the branch's auto-version run had failed on 2026-08-11), and registered the new `tests/unit/test_doc_scoring.py` in main's scoped-CI registry (`scripts/test_scope.py`), which did not exist when the branch was cut.

One statement below is now stale, superseded by the program of record (§2, corrected 2026-08-15): `publish-core.yml` is live and version-triggered — merging this PR (binary 0.4.0 + packaging 0.4.0) **will publish `core-v0.4.0` wheels to PyPI**. That wheel reaches only the opt-in `[core]` extra behind the always-complete Python fallback; the product's own release path is untouched.

## Test plan

- `make test` — 13,995 passed after rebase (unit + integration + contract; the 3 failures in the campaign sandbox reproduce identically on `origin/main` and are artifacts of the sandbox lacking the `gh` CLI / carrying a live session id — CI runners have both conditions inverted, and main's CI is green)
- `make lint`, `make go-test`, `make go-lint` — clean
- `make parity` — 36/36 (agentwatch + standup + analysis code + analysis docs), zero skips
- E2E: `run_doc_quality` with the sidecar discovered logs `analysis.score_docs served by the sidecar` and its signal + blob are **byte-identical** to a `YEABOI_GO=0` run
- Independent deep-tier review before the original push: one should-fix + five nits, all six addressed

~~Per the standing decision, **no yeaboi-core PyPI publish and no `core-v*` tag** until the full migration lands — devs and CI build the sidecar locally; end users keep the Python path.~~ *(superseded — see rebase note)*

Closes YEA-78


dinho149:
Both round-1 findings fixed in 63f346e:

**1. Cache key has no sidecar component** — took the review's second option: the convention is now written on the one lever that exists. `_DOC_SCORING_VERSION`'s comment states that a Go-side scoring fix with no Python change must bump it (the key carries no engine/binary component, so nothing else invalidates rows a divergent binary wrote — including after `YEABOI_GO=0`), and the Go twin's header in `go/internal/analysis/doc_quality.go` carries the mirror of that warning, since that file is where such a fix would land. Folding the binary version into the key was the alternative; it would split the cache between the two engine paths and churn every row on each sidecar upgrade, which is a behaviour change this rescue PR shouldn't smuggle in — the convention gets the same protection at the cost the review priced in.

**2. `zip(strict=True)` inside a best-effort writer** — replaced with the suggested explicit length check: a mismatch now logs one `warning` and skips the batch instead of raising through `run_doc_quality`'s blanket handler and zeroing the Documentation component. The mis-keying protection is unchanged (nothing is written on a mismatch). Regression test added: `test_cache_write_mismatch_skips_batch_instead_of_raising` (no raise, warning logged, nothing cached).

The nit is folded in too: CLAUDE.md's dual-maintenance row now names `team_learning.py`'s `_insight_item`/`_INSIGHT_MAX_ITEMS` as ported.

Verified: `make go-test`, `make go-lint`, `make parity` (36/36 unskipped), `make lint`, `make test` — all green.



---

PR #235, opened 2026-08-11
reach github without the gh cli so cd-deploy stops no-opping

dinho149:
## Summary

`cron/cd-deploy.md` is the safety net that carries a merged `cowork/` change into the live routine fleet. It has been failing on every firing, always at step 3, and the fleet has not been reconciled since.

The cause was not drift. `scripts/cowork_setup.py` reached GitHub **only** by shelling out to `gh`. The cloud routine session that runs the routine has no `gh` binary but does have `GH_TOKEN`, so `gh_ready()` took its "not on PATH" branch, `strict_exit()` returned 1, and step 3's stop condition — *"report and stop — do not proceed to the routines with a half-applied repo"* — halted the run before step 4, the reconciliation the routine exists to do. Every automatic deploy was a clean no-op while reporting a clean repo.

**Two transports, one set of operations.** `gh` is still preferred and its path is unchanged; when `gh` is missing or logged out, the same eight operations go over the REST API with `GH_TOKEN`/`GITHUB_TOKEN` through stdlib `urllib` (the script still imports nothing from `src/yeaboi` and nothing off PyPI). Two details the CLI hid: `gh variable set` is one verb but a POST to create and a PATCH to update (the wrong one is a 409, not a no-op), and `gh label list --limit 200` was one oversized page where REST caps at 100. The repo slug resolves from `gh`, then `GITHUB_REPOSITORY`, then the `origin` remote — the last is what makes it work in a routine session, since step 1 of the routine runs `git fetch`.

**Step 3 no longer stops the deploy.** The routine now reads the exit code: **2** means `cowork/` disagrees with itself and still stops; **1** means a GitHub write degraded — noted, carried into the Slack post, and the run continues. A label has no bearing on a trigger body, which is built from the `cowork/` files step 2 just validated. To make that rule mechanical rather than inferred from prose, the repo-disagreement path returns 2 (it previously shared 1 with the degradation case).

**Step 2 drops to `--check --local`.** Caught in review: the full `make cowork-check` now reaches GitHub too, so a run deploying a *new* workstream or a changed model id would abort on a missing label one step before step 3 creates it — the same halted deploy, moved earlier. The `pr-feedback` merge-gate probe is still reported, from the apply step.

A side effect worth noting: cloud runs of the doctor previously reported "clean" while silently skipping every GitHub check. They now actually check.

The branch also carries one earlier commit, `53856e3` "record the two newly registered routine urls" — an unrelated `cowork/README.md` URL-column edit that was already on this branch.

## Test plan

- `make test` — 11363 passed, 47 skipped
- `make lint`, `ruff check scripts/`, `make security` — all clean
- **The production shape, reproduced locally** — `env PATH=/usr/bin:/bin GH_TOKEN=… scripts/cowork_setup.py --strict`: exit 0, labels and all four variables applied over REST, merge gate read. Previously exit 1.
- No `gh` and no token: degrades with both remedies named, exit 1 under `--strict`, 0 without.
- `gh` present: byte-identical behaviour, `make cowork-check` still clean.
- ~45 new tests. The case that was actually failing in production — no `gh`, no token, under `--strict` — had no test at all, which is why it ran that way for as long as it did.
- A module-wide `_no_live_github` fixture now clears both token variables, resets the memoised transport state, and makes `_api` refuse by default. Without it, a developer with `GH_TOKEN` exported and no `gh` installed would have had the teardown tests resolve their own repository from `origin` and issue real label DELETEs — found in review, not hypothetical.

## Review findings

An independent `code-reviewer` pass raised 2 blockers, 3 should-fixes and 3 nits. **All eight are fixed in this branch**, including a crash it found that I had introduced and my own verification had missed: `merge_gate_armed()` keyed on the resolved transport could reach `gh api` on a machine with no `gh`, tracebacking `make cowork-setup` on a fresh clone. `_gh()` now reports a missing binary as exit 127 rather than raising.

## Definition of Done

Items 8 (Notion) and 9 (Slack) are handled by `pr-merged-close-loop` on merge. Item 10 (review feedback) is outstanding by definition — `claude-review.yml` has not run yet.

Closes YEA-81

### DIFF
--- a/cowork/README.md
+++ b/cowork/README.md
@@ -155,10 +155,10 @@ Cadence is tiered to surface size — a 1.2k-LOC mode asked for findings weekly 
 |---|---|---|---|---|
 | `cron/marketing-weekly.md` | `0 8 * * 6` Sat | marketing | `deep` | https://claude.ai/code/routines/trig_011f1J2fUGPhDQKSmjEMEiGs |
 | `cron/agents-standup.md` | `15 6 * * 1-5` weekdays | agents | `fast` | https://claude.ai/code/routines/trig_013tsooGjdnEMLRQcm7ZKU57 |
-| `cron/shipped-standup.md` | `0 18 * * *` daily | — | `standard` |  |
+| `cron/shipped-standup.md` | `0 18 * * *` daily | — | `standard` | https://claude.ai/code/routines/trig_0118jEhPuaKrCaUWCYQtVgEv |
 | `cron/digest.md` | `15 8 * * *` | — | `standard` | https://claude.ai/code/routines/trig_01VY1hbAZKeGuKA1GLyVhbow |
 | `cron/slack-relay.md` | `0 7-23 * * *` hourly | — | `fast` | https://claude.ai/code/routines/trig_01X18LBBBZ1FWEtx2Cmffyow |
-| `cron/release-promote-ask.md` | `0 9 * * 1` Mon | — | `fast` |  |
+| `cron/release-promote-ask.md` | `0 9 * * 1` Mon | — | `fast` | https://claude.ai/code/routines/trig_01G4TuU1wYY7GXJ1cEXZUNSu |
 | `cron/cd-deploy.md` | `0 4 * * *` daily + push (any branch) | — | `standard` | https://claude.ai/code/routines/trig_01AkW6ojpjKcra8H64R3Astr |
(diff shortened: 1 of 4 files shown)


dinho149:
Closing to reset the cowork backlog while the fleet is throttled back to a single lane. The branch `cowork-deploy` is **kept** — reopen this PR to fold the work into a later release batch.

dinho149:
Closing to reset the cowork backlog while the fleet is throttled back to a single lane. The branch `cowork-deploy` is **kept** — reopen this PR to fold the work into a later release batch.

Issue #228, opened 2026-08-11
[bug][integrations] confluence version-history enrichment truncates at 50 and reports complete


dinho149:
`_fetch_version_history` issues a single `rest/api/content/{id}/version?limit=50`, takes `results` once, and passes no `start` and follows no `_links.next` (`confluence.py:589-592`). On a page with more than 50 revisions the in-window editors can fall entirely outside that slice, so `_version_editor_items` yields nothing — and the caller still does `enriched_pages += 1` unconditionally (`:926`), so `incomplete_pages = len(version_candidates) - enriched_pages` stays 0 (`:929-937`), `on_partial` is never called, and `ConfluenceDiscoveryResult.complete` stays True.

**Impact** 4 · **Effort** M · **Risk** med

---

**What** — Version-history enrichment reads at most the 50 most recent revisions per page, silently, and the completeness-tracking machinery built for exactly this scenario is never triggered because the enrichment counter increments regardless of whether the fetch actually reached the target window.

**Why it matters** — A co-editor's standup contribution silently disappears and coverage reports the page as fully enriched. This is precisely the class the charter warns about — "a source that fails silently reads as a zero, and a zero is a number someone will believe." The contrast is inside one file: the discovery loop 100 lines above (`:712-789`) goes to considerable trouble to follow `_links.next`, retry on repeated ids via `_cql_next_with_retry` (`:496-540`), and report an `incomplete_reason` — so the honesty machinery exists in this same module and is simply not wired to this call site.

**Evidence** — `confluence.py:589-592` (single-page fetch, no pagination), `:926` (unconditional `enriched_pages += 1`), `:929-937` (`incomplete_pages` computation that can never go positive from this path), contrasted with `:712-789` and `:496-540` (the discovery loop's real pagination + retry + `incomplete_reason` reporting).

**Paths** — `src/yeaboi/tools/confluence.py` (`_fetch_version_history`, the enrichment loop around `:926`, `ConfluenceDiscoveryResult`).

**Related** — Explicitly distinct from #141: that one is Jira, four list calls in `tools/jira.py`, and truncation there is at least visible in the returned count. Here the truncation is *reported as complete* through the `ConfluenceDiscoveryResult`/`on_partial` channel that exists specifically to be honest about partial coverage. Linking as related, not duplicate.

[...]

Issue #214, opened 2026-08-10
[bug][standup] full-suite test run can hang on an unbounded live Jira connection attempt


dinho149:
While running `make test` for an unrelated PR, two full-suite runs hung for 30+ minutes inside `tests/unit/test_standup_engine.py::TestRunStandup::test_records_run_to_history` — a test that passes in ~1s when run alone. A fault-handler dump during the hang showed the test path making a real, live Jira connection attempt with no timeout bound. It's order-dependent (a third run did not reproduce the hang), pointing to test isolation leaking config/env state between tests rather than every run hitting the network.

**Impact** 3 · **Effort** M · **Risk** low

---

**What** — `test_records_run_to_history` can, depending on run order, drive a real (unmocked) Jira client construction with no connect/timeout bound, hanging the whole suite for 30+ minutes instead of failing fast.

**Why it matters** — A test suite that can silently hang for 30+ minutes on live network I/O, with no timeout, is a reliability problem for anyone running `make test` locally or in CI — it's indistinguishable from a genuine deadlock until someone waits it out. This is squarely `standup`'s test suite and its `roster.py`/`sprint_context` mocking, not `performance`'s (found incidentally while running `make test` on an unrelated `performance` branch).

**Evidence** — Hang observed running `make test` on an unrelated `performance` branch; fault-handler stack trace during the hang:
```
standup/engine.py:1436 → standup/roster.py:58 → performance/roster.py:20 → team_roster.py:225 → tools/jira.py:_make_jira_client (jira.ResilientSession retry loop)
```
Non-reproducing on a third isolated run of the same test alone.

**Paths** — likely `tests/unit/test_standup_engine.py` (missing/leaking mock for the Jira-client construction path) and/or `src/yeaboi/standup/roster.py` / `src/yeaboi/tools/jira.py` (no connect/timeout bound on `_make_jira_client`, which is what lets one leaked live attempt hang instead of failing fast).

Approve by adding the `claude-implement` label. Reject by closing this issue.

dinho149:
**Reclassified in place** — `cowork:queued`.

This find is on the auto-lane allowlist (`cowork/house-rules.md`, **The auto lane**), so it is not a
question for a human. **Nothing above has changed.** The `standup` sweep will pick it up,
re-check it against the allowlist, and open a PR; this issue closes when that PR merges. It no
longer counts against the workstream's proposal cap, and the daily digest no longer lists it.

[...]

PR #243, opened 2026-08-12
paginate notion_read_page so a long page is not silently halved

dinho149:
Closes YEA-87

`notion_read_page` called `client.blocks.children.list(page_id, page_size=100)` and read only `results`. Notion returns at most 100 children per request and signals the rest via `has_more`/`next_cursor` — both were ignored. Any page past 100 top-level blocks was silently halved, and the output carried **no** truncation marker, because the only existing notice fires on an 8,000-character clip that a 100-block prefix usually does not reach.

This is the tool the agent uses to read architecture docs, ADRs and runbooks, so the LLM planned against a document it believed it had read in full. A silent partial read is worse than a failed one: nothing downstream can tell.

The walk now follows the cursor to exhaustion, matching the idiom already used twice in this repo against the same SDK — `notion_read_page_text._children` in this file, and `_ensure_notion_brand_parent` in `export_targets.py`. `notion_read_page` was the last `blocks.children.list` call that did not paginate.

## Every partial read is now reported

Exactly one exit means "this is the whole page" (`has_more` false). All three "we cannot continue" exits set `capped`, which routes through the **existing** `[Truncated …]` suffix — no new or changed user-facing wording:

- an undocumented non-dict response mid-walk,
- `has_more: true` with a null or empty `next_cursor`,
- either ceiling: the 8,000-character content limit, or a new `_MAX_BLOCK_REQUESTS = 50`.

The request ceiling is not redundant with the character one. `_blocks_to_text` renders only `_TEXT_BLOCK_TYPES`, so a page built of images, dividers, embeds or column blocks renders to `""` no matter how many blocks it holds — the character ceiling never trips and the walk would follow cursors indefinitely, turning one call into hundreds of sequential requests against a ~3 req/s rate limit inside a single tool invocation. 8,000 characters cannot need more than a few dozen responses of 100 text-bearing blocks, so 50 is out of reach for a legitimate page.

## Regression evidence (auto-lane admission ticket)

Captured on a clean tree with `notion.py` reverted to the base commit and the new tests present:

```
>       assert "SECOND-PAGE-MARKER" in result
E       AssertionError: assert 'SECOND-PAGE-MARKER' in '=== Runbook ===\nURL: …\n\nblock 0\n…\nblock 99'
tests/unit/tools/test_tools_notion.py:285: AssertionError
1 failed, 1 passed
```

The output ends at `block 99` — the first response's 100 blocks, nothing past the cursor. After the fix:

```
59 passed, 1 warning in 0.80s
```

## Review

An independent `code-reviewer` read the diff before this PR opened and raised three `should-fix` findings, all fixed in `26e02b7`: the walk was not actually bounded (measured at 401 requests and still going on a media-block page); two exits still dropped content with `truncated=False`; and the `capped` flag was pinned by no test — deleting it left the suite green.

That last one is now verified by mutation. With `capped or` removed from the `truncated` expression:

```
FAILED …::test_stops_at_the_request_ceiling_and_reports_it
FAILED …::test_reports_truncation_when_has_more_carries_no_cursor
FAILED …::test_reports_truncation_when_a_later_response_is_malformed
3 failed, 56 passed
```

Previously that mutation left 56 passing and nothing failing.

## Find ranking

`impact 4 / effort S / risk low` — the highest-impact `auto` find of the run and the only one above impact 2. Three other `auto` finds were passed over (github top-level files rendered as directories, an unclamped Notion `page_size`, a github cache-key chore), all impact 2; they will be re-ranked next run.

Came out of the Edge-axis maintenance sweep that `cron/integrations-campaign.md` step 5 runs when no provider campaign is approved. The sweep rotated to GitHub and Notion, neither previously swept. Same *shape* as the open confluence issues #228/#230 but in a different provider's client, so a separate defect rather than a duplicate.

## Verification

- `uv run pytest tests/unit/tools/test_tools_notion.py -q` → **59 passed**
- `make lint` → **All checks passed!**
- Scope: two files, both inside the integrations charter's `Owns`. No `Extends` site, no `ui/mode_select/__init__.py`, no dependency added, no user-facing wording changed.
- `make parity` N/A — `tools/notion.py` has no Go twin.

**The full `make test` was not run to completion locally.** It hangs part-way through `tests/unit/test_standup_collector.py` in this environment — reproduced on a clean checkout of `main` with zero changes, which is open issue #214. CI gates the full suite here.

The five `tests/contract/test_notion_contract.py` tests fail in this sandbox with proxy 403s, **including ones this diff does not touch**; verified identical on the base commit `30bd3d7`, so environmental rather than caused by this change.

## Follow-up, deliberately not in this PR

`notion_read_page_text._children` (same file) has the identical unbounded-walk shape this fix just bounded, with a larger blast radius since it also queues nested children. Left alone rather than growing the change; it is a separate find for a later sweep.

---
_Generated by [Claude Code](https://claude.ai/code/session_011r6sanTQ6bUtrTSKcM4xmT)_


dinho149:
Both `should-fix` findings are fixed in `8596aae`. Both were real, and the first was a regression this PR introduced — worth saying plainly, since it is the kind of thing pagination changes cause and nothing else here would have caught it.

**Mid-walk `APIResponseError` discarding the read.** Correct, and the reasoning about why it only became reachable with this PR is exactly right. The `blocks.children.list` call now has its own `except APIResponseError` that keeps the prefix and reports it partial — but only once `blocks` is non-empty, so a failure on the *first* request still surfaces as an error rather than being dressed up as a truncated read of an empty page. That also makes the HTTP-error path agree with the malformed-response path, which was the inconsistency you named.

**The anomalous exits logging nothing visible.** Also correct: `get_log_level()` falls back to `WARNING` (`config.py:1009`), so the closing `logger.debug` records these for nobody. All three now `logger.warning` with the page id and the block count so far — the request ceiling included, on your reasoning that if it ever fires on a real page the ceiling is the thing wanting re-tuning.

Three tests added, and I checked they fail against the pre-fix code rather than assuming:

```
FAILED …::test_keeps_the_prefix_when_a_later_request_errors
FAILED …::test_every_partial_read_warns
2 failed, 1 passed
```

The one that passes both ways is `test_first_request_error_is_still_an_error` — deliberate, it guards against over-fixing this into swallowing a first-request failure.

`62 passed` on the file, `make lint` clean.

---

On the two nits, neither actioned, for different reasons:

**The `[Truncated at 8000 characters]` marker on an empty body** — agreed, and it is worse than cosmetic: an LLM reading that above an empty body is told something false about *why* the body is empty. Not changed here because the string is user-facing wording and this is an unattended auto-lane PR. Carrying it to the follow-up ticket alongside `notion_read_page_text._children`, which has the same unbounded-walk shape this PR just bounded.

**The `changelog_data.json` area tags** — the substance is right, and I verified it rather than taking it on trust: `standup/collector.py:585` uses `notion_recent_pages`, while `notion_read_page` is called from `agent/nodes.py:2435`, the planning intake. So two of the three 3.8.1 highlights are tagged `standup` for a fix that touches no standup path; `planning` is the accurate tag. (Small correction: the cited `:8-30` lands on the 3.8.0 entry, which is PR #240's standup work and legitimately tagged — the misattribution is in the 3.8.1 entry below it.)

Leaving it, and flagging rather than fixing it, because `changelog_data.json` is outside the integrations charter's `Owns` and the entry is user-facing copy. A machine editing another machine's changelog prose on an unattended PR is the specific thing the path boundary exists to prevent, and a wrong chip is not worth spending that. It is a two-word edit for whoever merges, or a `platform` follow-up.

---

dinho149:
Fixed in `cccbceb`, even though the round cap had already released the gate — this is the same regression class the PR exists to close, the change is small and verified, and `integrations` is at 0 proposal slots so there was nowhere to defer it to.

**The finding is correct, and the class hierarchy confirms it exactly.** I checked the installed SDK rather than reasoning from the names:

```
NotionClientErrorBase <- Exception
  RequestTimeoutError <- NotionClientErrorBase
  HTTPResponseError <- NotionClientErrorBase
    UnknownHTTPResponseError <- HTTPResponseError
    APIResponseError <- HTTPResponseError
```

`RequestTimeoutError` is a *sibling* of `HTTPResponseError`, so `except APIResponseError` never saw it, and `UnknownHTTPResponseError` — the gateway HTML 502 — is a sibling of the class I caught. Both fell to the function-level `except Exception` and discarded the prefix. Your point that the timeout is the more reachable half is right: it needs one slow response out of up to fifty back-to-back calls, not a quota decision.

Caught `NotionClientErrorBase` rather than bare `Exception`. The `if not blocks: raise` guard does make `Exception` safe on the first request as you say, but past that a `TypeError` in the loop would become a plausible-looking truncated read instead of an error — the SDK base is the same coverage without that.

**Nit taken and fixed, because it was the same standard this PR argued for elsewhere.** You were right that `test_every_partial_read_warns` pinned one of four exits while its name claimed all of them. Split into four tests, one per exit, each asserting its own message. Mutation-checked: with `logger.warning` → `logger.debug` in that function,

```
FAILED …::test_no_cursor_exit_warns
FAILED …::test_malformed_response_exit_warns
FAILED …::test_request_ceiling_exit_warns
FAILED …::test_mid_walk_failure_exit_warns
4 failed, 62 deselected
```

Previously that mutation failed nothing. The new timeout test is pinned the same way — reverted to `except APIResponseError`, `test_keeps_the_prefix_when_a_later_request_times_out` fails.

`66 passed`, `make lint` clean.

**Second nit — the O(n²) `_blocks_to_text` re-render — noted, not changed**, which matches your own read of it: bounded at 50×100 and dominated by the network round trips. It is recorded here so it surfaces if `_MAX_BLOCK_REQUESTS` ever moves.

---

dinho149:
Accounting for both review rounds, which the earlier replies described in prose but never recorded in the marker the gate reads. Nothing about the findings changed — this comment is the missing record, not a new claim.

**Round 1 — 2 findings, both fixed in `8596aae`.**

- *Mid-walk `APIResponseError` discarding the whole read.* The `blocks.children.list` call got its own `except`, keeping the prefix and reporting it partial, guarded by `if not blocks: raise` so a failure on the first request still surfaces as an error. That also made the HTTP-error path agree with the malformed-response path, which was the inconsistency the finding named.
- *The anomalous exits logging nothing visible.* `get_log_level()` falls back to `WARNING`, so the closing `logger.debug` recorded these for nobody. All three exits — including the request ceiling — now `logger.warning` with the page id and the block count so far.

**Round 2 — 1 finding, fixed in `cccbceb`.**

- *The mid-walk handler caught `APIResponseError` only, so a timeout still discarded the read.* `RequestTimeoutError` is a sibling of `HTTPResponseError`, not a child, so it never hit that handler. The `except` is now `NotionClientErrorBase` — the SDK base rather than bare `Exception`, so a `TypeError` in the loop still surfaces as an error instead of a plausible-looking truncated read. Verified present at `src/yeaboi/tools/notion.py:290` in the current head.

**Why the gate re-opened.** It had been released by the round cap, not by an account. Today's push (`b381fda`) merged `main` in to clear a merge conflict, which re-ran the check and surfaced the round-1 round as unaccounted — correctly, since the reply that fixed it carried no marker.

**What today's push actually contains.** Only the conflict resolution: both sides had prepended an auto-version entry. Took `main`'s `3.11.0` and re-bumped to `3.11.1` for this patch, and spliced this PR's changelog entry onto `main`'s list textually so the file is purely additive against `main` (19 insertions, 0 deletions) rather than reserialized. No change to `tools/notion.py` or its tests. `tests/unit/tools/test_tools_notion.py`, `test_release_channel.py`, `test_changelog.py`, `test_changelog_screen.py`, `test_version_row.py`, `test_bump_version.py` → 200 passed; `make lint` clean.

The one nit left unactioned is unchanged and still recorded above: the `[Truncated at 8000 characters]` marker on an empty body is user-facing wording, which the unattended auto lane may not change.



---

PR #250, opened 2026-08-12
split the pr-feedback gate into two lanes and scope ci to the change

dinho149:
## Summary

Three things, all about CI enforcing the right thing on the right lane and running
only what a change touches.

**1. The gate had been arguing with the wrong people.** A local branch got the same
two adversarial review rounds and the same blocking arithmetic as an unattended
cowork PR — but the reason the gate exists is that an unattended PR has nobody on
the other end to weigh a finding against the cost of not merging. A human's own
branch has exactly that person, already reading the review. `classify()` now
short-circuits there: the status is always `success`, carrying the finding count as
advisory, with the items still listed in the sticky comment.

Always `success` and never `pending` is the load-bearing half. A required context
sitting `pending` blocks a merge exactly as hard as a red one, so the human lane
short-circuits ahead of `waiting_reason()` entirely and cannot be pending by any
route. That is the precondition for making `pr-feedback` a required check at all —
which it has never been, despite five files describing it as the thing that enforces
DoD item 10.

On the lane that does enforce, only the first findings-bearing round blocks
(`BLOCKING_ROUNDS=1`). An adversarial review of a large diff finds something every
time, so running the loop to zero is not a condition that reliably arrives — four
consecutive rounds on #222 each produced real findings and none was a reason it
should not have merged. The exception is severity: the verdict marker gained
`critical=M`, the blocker-severity subset of `N`, and a later verdict carrying one
holds the merge however many rounds have run. That is what makes one blocking round
safe rather than merely cheap — the round that catches the blocker introduced *by*
the first round's fix is the one worth waiting for.

**2. A fix now has to say what it did.** "Push, and the next verdict reads `open=0`"
was the documented happy path, and it is right when a person did the fixing and is
about to click merge. When an agent does it, the whole record of what happened to
three findings is a number going down, and a subtraction cannot be reviewed.
`unaccounted_rounds()` holds the merge until a reply says what changed;
`silently_resolved()` does the same for a review thread the PR's author resolved
without ever answering in it. Both were conventions living in an agent prompt until
the lane started merging itself.

The response marker splits in two for this:

```
<!-- addressed: claude-review fixed=2 answered=1 -->
```

`fixed=` is a claim of work and is **accepted from the PR's own author**, because the
reviewer's next read of the diff checks it. `answered=` is a dismissal and stays
**refused from that author**, because nothing checks a disagreement. So a machine may
fix a finding and never dismiss one. `claude-review.yml`'s prompt is amended to
match — left alone it would have read a claimed fix as an answered finding, turning
the marker into a way to switch the reviewer off.

A `review_ledger()` renders it — one line per round, what it found, who wrote back —
into the sticky comment, `--json`, and `make pr-feedback`.

**3. CI runs the tests a change touches.** The two test jobs overlapped almost
entirely: `Unit tests` ran the 11,000 unit tests and `Integration & contract tests`
ran them a second time on its way to the 572 that are actually additional. Nothing
was scoped anywhere — a one-line docs edit rebuilt the front-end bundles,
cross-compiled Go, and ran 22,000 tests — and the suite was fully serial.

`scripts/test_scope.py` maps changed paths onto areas, one per cowork workstream so
the fleet and CI share a vocabulary, and a new `scope` job feeds the result to every
other job. Three rules keep it honest: `ALWAYS` runs the ~30 guards that scan the
repo rather than import a module, `GLOBAL` forces the whole suite for anything
everything reaches, and **any path the registry does not recognise runs everything**.
A selector's failure mode is silence, so the totality check is the feature —
`tests/unit/test_test_scope.py` fails the build when a source file is claimed by no
area or a test file is selected by nothing.

**The five required status checks never carry an `if:`.** GitHub counts a skipped job
as a passing required check only when the workflow ran and evaluated the condition; a
job skipped because its `needs:` was skipped posts no check at all and blocks the PR
forever. Scoping changes what they run, never whether they report — now asserted in
`test_workflow_schema.py`, including across the `needs:` chain.

Also picks up the five `tests/` root files (215 tests) that ran in no CI lane at all,
since `test-fast` and `test` both named subdirectories explicitly.

### Backward compatibility

Both marker changes fail in the safe direction. A verdict with no `critical=` reads as
`critical=0`, so anything already on an open PR keeps working and can only become
*less* blocking. A bare `<!-- addressed: claude-review -->` still means "all of them,
answered". An unrecognised field leaves the counts at 0, which under-claims.

### Still outstanding, deliberately

`pr-feedback` is **not yet** in the `main-branch` ruleset's required checks. Arming it
while `main` still has the old `classify()` would block every hand-shipped branch, so
that write happens after this merges. `make cowork-check` reports exactly this one
problem until then.

## Independent review

A fresh-context `code-reviewer` pass raised nine findings — none in the gate
logic, and all but the nits are fixed in the third commit. The ones worth naming:
a required check could be skipped by a *failing* `scope` job (a failed `needs:`
skips a dependent exactly as a skipped one does); `security`'s two
`sharing/*.py` entries were unreachable behind `artifacts-sharing`'s directory
claim, so `area_for` is most-specific-wins now; an `mcp/tools_*.py` change never
ran `test_mcp_server.py`; a rename reported only its destination, so a
cross-area `git mv` ran the wrong area's tests with the totality guard still
green; and a bare `<!-- addressed: … -->` from the PR's own author accounted for
a whole round, which is the contentless-comment hole the check exists to close.

### Residual risks, recorded rather than fixed

- **Cross-area imports are unprotected in both directions.** 100+ unit test
  files import a module owned by an area that does not select them —
  `src/yeaboi/prompts/standup.py` runs `tests/unit/prompts/*` but not
  `test_standup_engine.py`; `formatters.py` runs tui-ux only. The `integration`
  job used to backstop this by re-running the whole unit lane, and now runs
  `make test-slow`, so CI's unit coverage is whatever `scope` picked. That is
  the trade this PR is making. `Area.also` is the lever if a specific pairing
  burns, and `GLOBAL` is the bigger hammer.
- **Already-open unattended PRs go red on merge.** Any PR whose round-1 findings
  the reviewer has moved past acquires an `unaccounted_rounds` item the moment
  this lands. One comment clears it and the sticky says so, but the
  "every existing marker keeps meaning what it meant" story is about the
  markers — the gate itself is strictly stricter than before.

## Test plan

- `make test` — **12769 passed, 47 skipped** (298s), run against the rebased tree.
- `make lint` — clean. `make security` — SAST clean, `pip-audit` no known vulnerabilities.
- `actionlint .github/workflows/*.yml` — clean.
- `tests/unit/test_pr_feedback.py` is 187 tests (was 148). New coverage: a local-lane PR
  with `open=5` returns `success`; a local-lane PR with no review at all returns
  `success` and not `pending`; a cowork PR at round 1 blocks, at round 2 with
  `critical=0` passes and with `critical=1` blocks; a legacy marker parses as
  `critical=0`; a `fixed=` claim cannot dismiss a still-open finding; the PR's own
  author may write `fixed=` but not `answered=`; the account requirement is never
  capped; a bare marker still accounts for everything.
- `tests/unit/test_test_scope.py` — the two-way totality guard, plus the `ALWAYS` /
  `GLOBAL` / unknown-path-runs-everything rules.
- End-to-end eyeball against a hand-built snapshot: the red state produced three items
  (`producer`, `account`, `resolved`) with the ledger reading
  `Round 1 — 3 findings, 1 critical → **nothing written back yet**`; after the reply
  and thread comment it went `success` with `Round 1 — … → 2 fixed by @cowork-bot`.
- **This PR is itself the end-to-end test of the human lane** — expect exactly one
  Claude Review comment and a green `pr-feedback` whatever it finds.

## Definition of Done

- Item 7 (web bundles) does not apply — nothing under `frontend/` changed.
- Item 5 (surface parity) does not apply — this is CI and gate infrastructure, no new
  user-facing capability, so no `CAPABILITIES` row or `FeatureTip`.

Closes YEA-90


dinho149:
Round 1 — 2 findings, both **fixed** in bf9d1d05.

- **1. `account_for` — `answered=N` from the PR's own author cleared a round.** Fixed. You are right that this is the bare case with a number typed after it: the docstring justified the refusal as "claims no work, so nothing for the next review pass to check", and `fixed=0 answered=3` claims exactly as little. The test is `response.fixed == 0` now rather than `response.bare`. One claimed fix still admits the whole reply, because the reviewer's re-read of the diff is then a real check on it. Pinned by `TestADismissalIsNeverAnAccount` — including that a *maintainer* may still answer a whole round with no fix, since the refusal is about the applicant holding the key, not about the shape of the marker.

- **2. Local lane withdrew the human reviewer too.** Fixed, taking your second option rather than documenting the gap. Only `producer` items go advisory now; `thread`, `changes-requested` and `truncated` block on both lanes. Your framing is the reason — "nobody on the other end to weigh a finding" is the entire argument for the lane, and it is false by construction when the finding is a person's. The two symptoms you named are both gone: `definition-of-done.md` no longer promises something it had stopped doing, and the advisory sticky can no longer print "nothing here has to be answered before you merge" above an open thread, because that branch is now unreachable when one exists. Two tests that asserted the old behaviour were rewritten rather than deleted, with the reasoning in their docstrings.

  The load-bearing property survives and is pinned separately: a local branch may now be red for a person, but still cannot be `pending` for a machine that never spoke — which is the precondition for arming the ruleset.

On the four non-blocking observations, no reply needed but for the record: `Scope.slow`/`slow_paths` being wired-but-unread is real and deliberate for now (the slow lane is all-or-nothing, and `test_the_slow_lane_is_all_or_nothing` says so); the fallback's `tests/*.py` vs `FULL_UNIT`'s `tests/test_*.py` collects the same zero extra tests, and I have left the key-only comparison rather than widen it into a values check that would drift the other way; and the unreachable `after round {rounds}` branch is intentional headroom for `BLOCKING_ROUNDS` moving.

dinho149:
Re-posting the acknowledgement in the shape the parser currently on `main` understands.

Both findings are fixed in bf9d1d05 — the detailed reply is [above](https://github.com/dinho149/yeaboi.ai/pull/250#issuecomment-5265457943) and stands unchanged. The marker there is the **counted** form this PR introduces:

```
<!-- addressed: claude-review fixed=2 -->
```

`main`'s `ACK_RE` is `<!--\s*addressed:\s*([a-z0-9][a-z0-9-]*)\s*-->`, which requires the marker to end right after the producer key — so ` fixed=2` makes it not match at all, and the gate read the findings as unanswered. The new parser accepts both shapes (a bare marker keeps meaning "all of them"), but that compatibility only runs in one direction, and the code doing the reading here is the base branch's, because `pr-feedback.yml` is `pull_request_target` and checks out `main` by design.

So this is a one-way transitional gap, not a defect, and it closes when this merges.

Issue #206, opened 2026-08-10
[bug][planning] Task.label / test_plan / ai_prompt silently dropped on session --resume


dinho149:
When a session is resumed, every task silently loses its label, its test plan, and its AI-assistant prompt — the last of which is the self-contained prompt meant to be pasted into an AI coding tool, a headline feature. Nothing warns the user, logs the loss, or fails a test; the values just quietly revert to defaults. Anyone relying on `--resume` to pick a session back up gets tasks that look intact but have silently lost real content.

**Impact** 8 · **Effort** S · **Risk** low

---

**What** — `_dict_to_task` only reconstructs `id`/`story_id`/`title`/`description` from the serialized dict, even though `Task` also carries defaulted fields `label` (`TaskLabel`), `test_plan`, and `ai_prompt`. `_serialize_state` writes the full `asdict()` of every `Task` (including those three fields), but `_dict_to_task` never reads them back out, so they silently revert to the dataclass defaults (`CODE`, `""`, `""`) on every `--resume`.
**Why it matters** — This is exactly the "silent session-restore corruption" the planning charter singles out as its top concern: a resumed session loses every task's label and AI-assistant prompt with no warning, log line, or test catching it.
**Evidence** — `src/yeaboi/sessions.py:347-354` (`_dict_to_task` omits `label`/`test_plan`/`ai_prompt`) vs `src/yeaboi/agent/state.py:227-249` (`Task` fields) and `src/yeaboi/sessions.py:220-227` (`_serialize_state` `asdict()`s the full `Task`, including the dropped fields). `tests/integration/test_sessions.py:285` (`_make_full_state`'s fixture `Task` never sets `label`/`test_plan`/`ai_prompt`, so the existing round-trip test at line 461 can't detect the loss); no test anywhere calls `_dict_to_task` directly (grep for `_dict_to_task` in `tests/` returns nothing).
**Paths** — `src/yeaboi/sessions.py`, `tests/integration/test_sessions.py`, `tests/unit/test_state.py`

Approve by adding the `claude-implement` label. Reject by closing this issue.

dinho149:
**Reclassified in place** — `cowork:queued`.

This find is on the auto-lane allowlist (`cowork/house-rules.md`, **The auto lane**), so it is not a
question for a human. **Nothing above has changed.** The `planning` sweep will pick it up,
re-check it against the allowlist, and open a PR; this issue closes when that PR merges. It no
[...]

Issue #140, opened 2026-08-04
[bug][integrations] jira cloud roster pagination crashes past page 1


dinho149:
Atlassian shut down the old Jira search API months ago, and one roster-fetching function still pages through it in a way that crashes on any Jira Cloud project with more than 100 matching issues. Real Cloud customers with sizeable projects hit this today. The recorded test suite only exercises the retired self-hosted flavour of Jira, so the tests stay green while the path everyone actually uses is broken.

**Effort** M

---

## What

Atlassian deprecated `GET /rest/api/2/search` (and `/rest/api/3/search`) on 2025-05-01 and fully shut it down by 2025-10-31 (Atlassian changelog CHANGE-2046). The pinned `jira` PyPI library (3.10.5, per `uv.lock`) works around this in `JIRA.search_issues`: when the client is Cloud (`self._is_cloud`) and `startAt == 0`, it transparently delegates to `enhanced_search_issues()`. But when Cloud and `startAt != 0`, it raises `JIRAError("The \`search\` API is deprecated in Jira Cloud. Use \`enhanced_search_issues\` method instead.")` with no fallback.

`jira_assignee_roster()` in `src/yeaboi/tools/jira.py` (~lines 961-1018) is the one function in the module that paginates manually with an incrementing `startAt` in a `while True` loop (`page_size = 100`). Against a real Jira Cloud project where more than 100 issues match the roster query, the second page request throws. The `except JIRAError` handler re-raises via `_jira_error_msg`, which has no case for this error (it lacks `status_code`), so it falls through to a generic branch that may itself be fragile since this `JIRAError` instance wasn't given a `response` (so `.text` may not behave as expected).

Separately: the contract-test fixture `_make_test_client()` in `tests/contract/test_jira_contract.py` passes `get_server_info=False`, which its own docstring says sets `_is_cloud=False` — the classic Server/DC code path. That means the entire Jira contract-test suite only exercises Server/DC, never Cloud (neither the `enhanced_search_issues` delegation nor the Cloud pagination failure above). Since Jira Server was retired years ago, the tests are green on a path almost nobody hits, while the path everybody hits is uncovered.

## Why now

[...]

PR #252, opened 2026-08-12
split the proposal queue so the fleet builds its backlog instead of asking about it

dinho149:
## Summary

Slack was full of "🗳️ Decisions — ✅ to approve" digests about **bugs**, and nothing shipped. That is
backwards: `house-rules.md` puts bugs, chores, docs and security in the unattended auto lane, and
`digest.md` already claimed those sections were "empty on most days". The doc was right about the
intent and wrong about the mechanism.

**The dedupe was lane-blind in both passes.** `sweep-procedure.md` step 4 and `cowork-scout.md` step
5 dropped any find restating an open proposal, whatever lane the find was in — so a bug the auto lane
was entitled to ship that morning was dropped because an unanswered *question* about the same bug was
open. Then the digest closed that question on the 14-day timer, which both passes read as a human's
rejection and suppressed the find permanently. 45 open proposals, none ever closed, against a cap of
2 per workstream that only a human answer releases: `platform` at 10 and `integrations` at 9 could
not file at all.

Three commits:

1. **`cowork:queued`** beside `cowork:proposal` — a work item waiting on the fleet, beside a question
   waiting on a human. Mutually exclusive, which is load-bearing rather than tidy: every consumer
   asks for `labels=cowork:proposal,…` and that parameter is AND-only, so exclusivity keeps
   `digest.md`'s three queries, `codeql-triage.yml` and `flaky-test-hunter.yml` correct with no edit.
   A queued item holds no slot. `--queued` / `make cowork-queue` report the queue in build order.
2. **Lane-aware dedupe.** Step 4 gets four outcomes instead of one; step 5 reads the queue first and
   merges it with today's finds into one comparator (critical → impact → risk → queued, oldest
   first). A queued issue is a hypothesis, not a permission: step 5 re-checks the allowlist, bounces
   what fails back to `cowork:proposal` naming the condition, and closes what no longer reproduces.
   `Closes #<n>` is the only thing that closes a queue entry.
3. **`--migrate-proposals`** backfills the existing backlog once — **36 queued, 8 held, 1 skipped**,
   so the digest asks about eight things instead of forty-five.

## Two bugs found by pointing the migration at real data

- **The `gh` branch of the issues read was unpaged.** It stopped at 30 of 45 and reported a *clean
  plan* over the fifteen it never saw. The REST branch has paged from the start; this one had not,
  and the identical hazard is already documented on `open_blocked_reports`. Now `--paginate` with
  `per_page=100`, asserted in both transport tests.
- **The evidence check matched one spelling.** The scribe writes `**Evidence**`, older issues write
  `## Evidence`, and #140 — which carries two `file:line` references under an H2 — was held as though
  it had none.

## The risk this creates, and its guards

Four structurally-empty type sections mean "nothing today" becomes the normal morning, so a stalled
fleet would look like a clean codebase. Guarded three ways: a 🛠️ **Queued** health section on every
morning the digest posts; two new *positive* stop conditions (an "approved, no PR yet" entry, or a
workstream queued-but-not-merging for 21 days, each forces a post); and a 🔇 **Silent** clause that
reports a stopped drain with its cause.

The queue is **drain-only** — nothing files into it, a passed-over `auto` find is still dropped — so
it is seeded once and only shrinks.

## Test plan

- `make test` — 11722 passed, 47 skipped
- `make lint` — clean
- `uv run python scripts/cowork_setup.py --check` — the merge-gate finding is gone; it now reports
  only that `cowork:queued` does not exist yet, naming `make cowork-setup` as the remedy
- `--migrate-proposals` dry run against the live 45-issue backlog: 36/8/1, and every held row was
  inspected by hand
- New: `TestTheQueueSplitsTheBacklog` (16), `TestTheQueueContract` (10 markdown-contract rules, each
  of which fails silently at run time), `TestMigrateProposals` (15), `TestMigrateProposalsApply` (6)

## Notes

Related to the same unblocking pass: `pr-feedback` is now a required status check on `main`
(ruleset 14008705), which is what lets `sweep-procedure.md` step 5 arm `--auto` at all — it had been
refusing to. PRs #184, #213 and #145 merged once refreshed, unblocking `platform`, `performance` and
`poker`.


dinho149:
Both findings were real and are fixed in `b3668936`. Neither was dismissed.

**1. The score line never parsed.** Confirmed against the live data before touching it:

```
'**Impact** 4 · **Effort** S · **Risk** low'  -> (None, None)
'**Impact** 5 · **Effort** M · **Risk** high' -> (None, None)
```

Exactly as described — `\d+` on a field the scribe writes as a word, and because both halves shared
one pattern the unmatched risk discarded the impact beside it. Now two independent searches, with
risk normalised to one 1–3 scale whichever way it was written (`low|med|medium|high|\d+`), and either
half may be `None` without taking the other with it. On the live backlog that is **27 of 45 rows
carrying a parsed impact, against none before**, and `queue_report` orders on real numbers instead of
degenerating to oldest-first.

The fixtures fed `**Risk** 1`, which is the whole reason CI was green over it. They use the scribe's
template verbatim now, plus a parametrised test over every spelling — including the numeric form a
hand-written issue may still use.

One thing worth flagging that came out of checking the fix: issue #206's body says `**Impact** 8`,
outside the scribe's documented 1–5 scale. The parser reads it faithfully rather than clamping — it
only affects sort order, and silently rewriting a filer's number seemed worse than letting an 8 sort
first. Nothing enforces that scale at write time, which is a scribe-side gap and not this PR's.

**2. The carve-out asymmetry.** Correct, and the consequence you named is the sharp end — a *second
public issue* re-asking a decision `triage-policy.yml` already records. Fixed in all three places
rather than one:

- `sweep-procedure.md` step 4 now refuses to reclassify a `codeql:` issue, with the dedupe breakage
  as the stated reason;
- `house-rules.md` **The queue** carries the same rule, and says plainly that this is the one
  consumer for which exclusivity is a hazard rather than a convenience — which is the claim in the
  PR body you were right to push back on;
- `codeql-triage.yml` now dedupes on `--label cowork --search`, which sees either label. Defence in
  depth: the prose stops a sweep doing it, and the widened search means the workflow survives it
  happening anyway.

`test_the_codeql_carve_out_exists_in_the_recurring_path_too` pins all three, on the reasoning that a
carve-out present in one door and not the other is a hole that opens a week later rather than never.

**Nits:** the bounce is `--add-label cowork:proposal --remove-label cowork:queued` now, matching
`_reclassify` and step 4 — you're right that one `gh issue edit` is a single PATCH, but the paragraph
directly above it tells an agent which order to use and it should not contradict itself. The
`--paginate` + `json.loads` concatenation hazard is real and predates this PR as you say; it fails
loud, so I have left it rather than widen the diff.

`make test` 11732 passed, `make lint` clean.

Issue #172, opened 2026-08-06
[security][web-ux] Third-party CDN script (unpkg lenis) loaded on all 20 docs/ pages with no SRI


dinho149:
Every page of the public yeaboi.ai marketing site pulls a third-party JavaScript library straight from a CDN with no tamper check. If that CDN or the package it serves were ever compromised, the injected script would run on the production domain with nothing on the page to catch it.

**Impact** 2 · **Effort** S · **Risk** low

---

**What** — All 20 pages under `docs/` load `<script src="https://unpkg.com/lenis@1.3.4/dist/lenis.min.js"></script>` with no `integrity` or `crossorigin` attribute. Adding a Subresource Integrity hash (and `crossorigin="anonymous"`) is a narrow, low-risk hardening — one generated hash reused across all pages.

**Why it matters** — Google Fonts and GA tags on these same pages are expected, out-of-scope external references for a marketing site, and this is unrelated to the `frontend/` bundle self-containment rule (that rule governs the app bundles, not the docs site). But an npm package fetched unpinned from a CDN and executed with no integrity check is a real supply-chain gap: an unpkg or lenis compromise would run arbitrary JS on the production domain with zero client-side detection.

**Evidence** — `docs/index.html:365` and `docs/docs/architecture.html:390` (per `grep -rln unpkg.com docs/`, all 20 site pages carry the same tag): `<script src="https://unpkg.com/lenis@1.3.4/dist/lenis.min.js"></script>` — no `integrity` attribute exists anywhere in `docs/`.

**Paths** — `docs/`

Note: this is a `web-ux` finding of type `security` — it's markup in `docs/`, not CSP policy/header construction, which belongs to the `security` workstream (`web/security.py`).

Approve by adding the `claude-implement` label. Reject by closing this issue.

dinho149:
approved via Slack ✅ by onoureldin — https://slack.com/archives/C0BMADQQN1Z/p1786263812848879

---
_Generated by [Claude Code](https://claude.ai/code)_

dinho149:
approved via Slack ✅ by onoureldin (onoureldin@gmail.com) — https://yeaboai.slack.com/archives/C0BMADQQN1Z/p1786263808.836259?thread_ts=1786263808.836259

---
_Generated by [Claude Code](https://claude.ai/code)_

dinho149:
[...]

PR #253, opened 2026-08-12
read claude-implement as a level, so an approval that produced nothing retries

dinho149:
## Summary

`claude-implement` is a **level** — "this issue is approved and unbuilt" — but `claude.yml`'s
implement job only ever observed its **edge** (`issues: labeled`). Everything that can come between
the two was unrecoverable:

- **a second ✅ is a silent no-op.** `gh issue edit --add-label` on a label that is already there
  emits no `labeled` event. Issue #172 was ✅'d in Slack five times between 2026-08-09 and 08-11 and
  built **once**; the relay acked all five, so nothing anywhere reported that four did nothing.
- **the one run that did fire produced nothing** — 231s, exit `success`, no branch, no PR, no
  comment — and nothing retried it.
- `claude.yml` has concluded `skipped` on **100 of its last 100 runs**.

## Three doors, none of them trusted

`workflow_dispatch` with an issue number, the existing label event, and a `<!-- implement-retry -->`
marker gated on `author_association`. A new **"Resolve the target issue"** first step re-reads the
issue's live state and labels and refuses anything not open and carrying `claude-implement` — so the
trigger says *which* issue and this step decides *whether*. A forged marker from a write-capable
account can only re-run something a human already approved.

It is also strictly better than the label path for #172's own failure: that run read its labels from
an event payload captured in the same second the relay stripped three of them, and built with **no
charter at all**. The tier step, the prompt and the guard now all read from the live values.

Plus a job-level `concurrency` group, which did not exist — five ✅s could in principle have started
five 110-turn agents on one issue.

## The reconciler

`implement-reconcile.yml`, every six hours, no model — `gh` and `jq`. For each open
`claude-implement` issue with no PR, no branch and no run in flight, dispatch the job. **Three
attempts**, counted from `<!-- implement-attempt -->` comments, then `implement-blocked` and stop: a
level-triggered re-firer with a broken implement job would otherwise burn a 110-turn agent every six
hours forever, which is the one way this can be worse than the bug it fixes. The digest reports that
state, since the issue keeps `claude-implement` and the age-out still refuses to touch it.

## Four holes in the outcome guard

Each one turned it back into a green tick:

1. **a missing execution log `::warning::`d and set denials to 0** — a guard that cannot check,
   passing. The path is hardcoded against one observed run of one pinned action version. Now an
   error, and it reads `steps.implement.outputs.execution_file` first. An `error_max_turns` subtype
   fails too — today that is only inferable from the branch-without-PR path.
2. **"a bot commented in the last two hours"** counted `codeql-triage`, `flaky-test-hunter`,
   `feedback-remediation` and `pr-merged-close-loop`, and would have counted the reconciler's own
   attempt marker. Now it requires `<!-- implement: ambiguous -->` — a marker the prompt is told to
   write — posted after this job started.
3. **the branch probe knew one convention.** `claude-code-action` names its own branches
   `claude/issue-N-…`, which read as "no branch". It asks `gh pr list --search "$N in:body"` first
   now — the authoritative question, and the same one `digest.md` asks.
4. **a failed guard went nowhere.** The job turned red, the issue kept its label, the age-out
   refused to touch it, and nothing re-fired. It now comments the run URL and applies
   `implement-blocked`.

## Test plan

- `make test` — 11691 passed, 47 skipped · `make lint` clean · `actionlint` clean on both workflows
- `test_claude_workflow.py` selected the guard by **position**, which held only while nothing ran
  after it — and the fix for a guard that dies silently is exactly to add a step after it. By name now.
- New relay tests replay the recorded 2026-08-09 thread both ways: `approved_check` False → the
  `approve` it was, True → the `refire` it should have been, None → still `approve` (the asymmetry
  with `is_promotion` is deliberate, and the docstring says why).
- The label-safety assertion was parametrised over every verb by name — it sampled a thread before,
  so it could not reach `refire` at all, and a safety check that silently skips the newest verb is
  not one.

## Note

Four approval tests were reaching the **real** `gh issue view` and passing only because they never
hit that seam. They stub it now — which is what `build_plan`'s docstring already said they must.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -9,6 +9,18 @@ on:
     types: [opened, assigned, labeled]
   pull_request_review:
     types: [submitted]
+  # The level-triggered door into `implement`, below. `claude-implement` is a
+  # *level* — "this is approved and unbuilt" — but `issues: labeled` only observes
+  # its *edge*, so an approval that arrives twice, or whose run produced nothing,
+  # had no way to fire again. Issue #172 was ✅'d in Slack five times between
+  # 2026-08-09 and 08-11; only the first became a `labeled` event, and this
+  # workflow concluded `skipped` on 100 of its next 100 runs.
+  workflow_dispatch:
+    inputs:
+      issue:
+        description: "issue number to (re-)implement — must already carry claude-implement"
+        required: true
+        type: string
 
 jobs:
   claude:
@@ -55,10 +67,21 @@ jobs:
   # claude-review.yml, and auto-version.yml like any human-authored branch.
   implement:
     if: |
-      github.event_name == 'issues' &&
-      github.event.action == 'labeled' &&
-      github.event.label.name == 'claude-implement'
+      (github.event_name == 'issues' &&
+       github.event.action == 'labeled' &&
+       github.event.label.name == 'claude-implement') ||
+      github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'issue_comment' &&
+       !github.event.issue.pull_request &&
+       contains(github.event.comment.body, '<!-- implement-retry -->') &&
+       contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association))
     runs-on: ubuntu-latest
+    # Five ✅s on one issue could otherwise have started five 110-turn agents on
+    # it. There was no concurrency key on this job at all; the only thing that
+    # stopped it was the edge trigger being broken.
+    concurrency:
+      group: implement-${{ github.event.issue.number || inputs.issue }}
+      cancel-in-progress: false
     permissions:
       contents: write
       pull-requests: write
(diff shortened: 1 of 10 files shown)


dinho149:
All three fixed in the latest commit. All three were real, and the framing — that they are the
*terminal* states, the ones nothing reports twice — is what made them worth catching.

**1. "Remove the label to let it retry" did the opposite.** Exactly as described: nothing deletes the
attempt markers, so clearing `implement-blocked` landed the issue straight back in the cap branch on
the next tick and re-posted the same notice every six hours. Following the documented recovery was
what generated the noise.

Attempts are now counted since the most recent `UnlabeledEvent` for that label
(`gh issue view --json timelineItems`), falling back to the lifetime count when there has never been
one. Clearing the label genuinely resets it, and the notice now says that is what clearing it does.

**2. `if: failure()` was too broad.** You're right that the guard directly above it is scoped with
care and the reporter was not — and that a `uv sync` 503 turning into an unrecoverable state is the
exact loss class this PR exists to close. The guard has `id: guard` now and the reporter is
`if: always() && steps.guard.conclusion == 'failure'`, so the label means what the step name says.

**3. `--state all` — the worst of the three,** because it silently disarms everything else on the
path this PR makes normal. A closed-unmerged PR from attempt 1 is precisely the state a re-fire
exists for, and it was answering "is there a PR" for attempt 2. Both call sites now select on
`state == "OPEN" or mergedAt != null`. Your note about `implement-reconcile.yml:69` was the sharper
half: there the issue is never re-fired *and* never blocked, so it stops existing to the fleet
entirely — `digest.md:339` asks the identical question and would also have seen it. Fixed in both,
and `test_a_closed_unmerged_pr_does_not_count_as_a_pr` is parametrised over both files, because they
ask the same question and a fix in one is not a fix in the other.

**Nits:** the `$GITHUB_OUTPUT` heredoc delimiter varies per run now — it wraps an issue title on a
public repo, and the human-approval gate in front of it is the real defence but this half is free.
Added `tests/unit/test_implement_reconcile.py` (13 tests) covering the invariants you named: the
attempt cap, the terminal label, the blocked-check-before-count *ordering*, both branch conventions,
and that the workflow can do nothing but dispatch — no `--add-label claude-implement`, no close, no
merge.

I left the `gh run list --status in_progress` imprecision as you suggested. You're right on all three
counts (dispatch runs carry no issue number in `displayTitle`, `queued` is missed, and the substring
is unanchored) — the concurrency group and the attempt cap bound it to wasted work rather than
damage, and tightening it would mean matching on run inputs, which is a bigger change than the
precision is worth here.

`make test` 11704 passed, `make lint` clean, `actionlint` clean on both workflows.

PR #254, opened 2026-08-12
give --migrate-proposals a gh write path, and stop tests reaching the real gh

dinho149:
## Summary

Two bugs, the second found while fixing the first.

**1. The migration could not write on the transport it read with.**

`_issues_labelled` asks REST on both transports deliberately (routine sessions' egress proxy refuses
GraphQL, so `gh issue list --json` is out). But `_reclassify` had only the REST branch. On a laptop
with `gh` logged in and no `GH_TOKEN` exported — the ordinary way a human runs this, and per the
`--strict` refusal the *only* way it is allowed to be run — all 45 issues read back fine and every
single write failed:

```
[cowork] note: #140 not reclassified: no GH_TOKEN or GITHUB_TOKEN in the environment
[cowork] note: #141 not reclassified: no GH_TOKEN or GITHUB_TOKEN in the environment
… ×36
```

It branches like every other writer in the file now, using `--add-label`/`--remove-label` — never
anything that could replace a label set.

**2. Finding that, the tests wrote to the real repository.**

`TestMigrateProposalsApply` stubbed `_api` and asserted on the calls. That was complete for exactly
as long as `_reclassify` had one branch. The moment it grew a `gh` branch — and `TRANSPORT` defaults
to `"gh"` — those same tests ran real commands against this repo: `gh issue edit 7 --add-label
cowork:queued` plus four `gh issue comment` calls, landing on a PR merged months ago. Undone by hand
(label removed, all four comments deleted, verified zero issues carry the label).

A test that asserts on writes is precisely the test that will *make* them when a seam moves, and
"stub the right seam" is not a property anything checks. So:

- **`_gh_transport` gets `_run`**, its single process seam, with **every** spawn in the file routed
  through it — `gh` and `git` alike, which is what makes replacing one name total. Named rather than
  reusing `subprocess.run` because that object is shared with every other module, which is the trap
  `conftest`'s `_no_real_package_install` already documents for the installer.
- **`_no_real_gh_calls`** blocks it, autouse, for argv starting with `gh`. Local `git remote get-url`
  delegates through: it cannot reach GitHub, and blocking it failed seven tests that had done nothing
  wrong — a guard with that much collateral gets loosened rather than kept. `BaseException`, like its
  sibling, so a broad `except Exception` in code whose contract is to degrade cannot swallow it.

Verified the guard catches the original call: `_reclassify` on the `gh` branch with nothing stubbed
now raises `RealGitHubWriteBlocked: test tried to spawn the real gh CLI: ['gh', 'issue', 'edit', …]`.

## Note on the fallout

Several tests stubbed `transport.subprocess.run` or `setup.subprocess.run` to control the
transport's spawn. Those worked only because both names resolve to the *same shared module object* —
an accidental coupling the `_run` seam correctly breaks. They stub `transport._run` now, which is
what they meant all along.

## Test plan

- `make test` — 11734 passed, 47 skipped · `make lint` clean
- New: `test_it_writes_over_gh_too_not_only_rest` and `test_the_gh_branch_adds_before_it_removes`,
  the latter pinning the same add-before-remove ordering the REST branch already had

### DIFF
--- a/scripts/_gh_transport.py
+++ b/scripts/_gh_transport.py
@@ -60,6 +60,20 @@ class ApiResult:
 
 # --- the CLI half ------------------------------------------------------------
 
+# This module's single process seam — EVERY spawn in this file goes through it,
+# `gh` and `git` alike, which is what makes replacing this one name total. Named
+# so it can be replaced without touching
+# the shared `subprocess` module — patching `subprocess.run` itself would patch it
+# for every other test in the suite, which is the trap `tests/conftest.py`'s
+# `_no_real_package_install` documents for the installer.
+#
+# It exists because the test suite reached the real `gh` once and wrote to the
+# real repository: a migration test stubbed the REST seam and asserted on the
+# calls, which was complete until the function under it grew a `gh` branch, and
+# then `gh issue edit 7 --add-label …` plus four comments landed on a merged PR.
+# `_no_real_gh_calls` blocks this name.
+_run = subprocess.run
+
 
 def gh(*args: str) -> subprocess.CompletedProcess[str]:
     """One `gh` call, reporting a missing binary the way `gh` reports a failure.
@@ -71,7 +85,7 @@ def gh(*args: str) -> subprocess.CompletedProcess[str]:
     traceback out of a script whose whole contract is to degrade with a remedy.
     """
     try:
-        return subprocess.run(["gh", *args], capture_output=True, text=True, check=False)
+        return _run(["gh", *args], capture_output=True, text=True, check=False)
     except OSError as error:
         return subprocess.CompletedProcess(["gh", *args], 127, "", str(error))
 
@@ -292,7 +306,7 @@ def _slug_from_remote(root: Path | None) -> str | None:
     """``owner/name`` parsed out of `git remote get-url origin`."""
     argv = ["git"] + (["-C", str(root)] if root else []) + ["remote", "get-url", "origin"]
     try:
-        result = subprocess.run(argv, capture_output=True, text=True, check=False)  # noqa: S603 - literal argv
+        result = _run(argv, capture_output=True, text=True, check=False)  # noqa: S603 - literal argv
     except OSError:
         return None
     if result.returncode != 0:
--- a/scripts/beta_signoff.py
+++ b/scripts/beta_signoff.py
@@ -34,13 +34,13 @@ from __future__ import annotations
 import argparse
 import json
 import re
-import subprocess
 import sys
 from pathlib import Path
 
 ROOT = Path(__file__).resolve().parent.parent
 sys.path.insert(0, str(Path(__file__).resolve().parent))
 
+import _gh_transport as transport  # noqa: E402
 import cowork_relay  # noqa: E402
 import release_channel as channel  # noqa: E402
 import release_surfaces  # noqa: E402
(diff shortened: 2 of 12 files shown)


dinho149:
Both findings confirmed and fixed. Finding 1 was correct, and you were right to ask me to re-run the
proof — it was a false positive.

**1. The guard patched an object nothing called.** Verified before touching anything, with a throwaway
test run across multiple files:

```
cowork_setup.transport         id=4384906880
sys.modules['_gh_transport']   id=4386531776
same object: False
AssertionError: the guard patches a DIFFERENT module object than cowork_setup calls
```

Two distinct module objects from the same path, exactly as you described — and the reason my stated
verification passed is exactly the one you named: run against `test_cowork_setup.py` alone,
`test_gh_transport.py` is never collected, the fresh module never exists, and the registry entry *is*
the object `cowork_setup` holds. So the guard's own proof was only ever exercising the case where the
bug cannot occur.

Fixed by patching every distinct transport reachable from the loaded scripts modules, deduped by
identity, rather than the `sys.modules` entry. I chose that over making `test_gh_transport.py` reuse
the registered module because it holds regardless of collection order — the fix at the source depends
on nobody reintroducing a path-loader later, and this repo has several.

**2. `beta_signoff` — the sharp one.** You're right that the fixture docstring read as a suite-wide
invariant it did not have, and right about which of the three sites matters: `_gh` there *writes*, and
its only protection was tests remembering to stub it, which is precisely the property the PR argues
nothing checks. All three go through the seam now.

Routing `cowork_setup`'s ruleset probe found a third live-call site you didn't have visibility into:
it was the one call in that file with its own `subprocess.run`, which made it the one call its tests
could not stub — `TestStrict` was issuing a live `gh api .../rules/branches/main` on every run. It uses
`_gh` now like every other CLI call in the file. The `api` fixture's teardown tests were doing the same
through `github_ready()` → `gh auth status`. Both surfaced only because the guard started refusing,
which is the guard working before it had a test of its own.

**Now it does have one.** `test_zz_gh_guard.py` asserts the reach in the multi-module condition rather
than the single-file one, checks every loaded transport carries the block, drives the write path that
caused the incident end to end, and **skips loudly if fewer than two transports are loaded** — so a
vacuous pass can never again be mistaken for a covered one. That last part is the direct lesson from
my bad verification.

**Nit:** correct, `raising=False` was unreachable — `real = transport._run` would have raised first.
Removed.

`make test` 11739 passed, `make lint` clean.

dinho149:
Accounting for round 1's two findings. Both were **fixed in code**; neither was argued away.

**1. The guard patched a module object nothing called.** Fixed in `882bbd31`. Verified the claim
first, because it contradicted a proof I had already put in the PR body — and the proof was the thing
that was wrong:

```
cowork_setup.transport         id=4384906880
sys.modules['_gh_transport']   id=4386531776
same object: False
```

`scripts/` is not a package, and two loaders disagree about who owns the name, so in a multi-file run
there are two module objects from one file. Running a single test file collapses them, which is why my
verification passed. The guard now patches every transport reachable from the loaded scripts modules,
deduped by identity.

**2. Three `gh` spawns outside the seam.** Fixed in the same commit. `beta_signoff._gh` was the one
that mattered — it *writes* (`gh issue comment`, `--add-label release:promote`) with `cwd=ROOT` — plus
`cowork_relay`'s issue read and `cowork_setup`'s ruleset probe. All three route through the seam now.
Routing the probe exposed that `TestStrict` had been issuing a live `gh api .../rules/branches/main`
on every run, and the `api` fixture's teardown tests a live `gh auth status`.

**Since round 1**, two CI failures on this branch — also fixed, and both the same root cause seen from
the other side: `test_zz_gh_guard.py` was selected by nothing under #250's scoped CI (now in `ALWAYS`),
and `test_cowork_relay.py` was making live `gh issue view` reads that only failed in a *scoped* run,
because which modules are loaded decides which transport objects the guard reaches.

`make test` 12932 passed · `make lint` clean · scoped shapes re-run individually to reproduce CI's
selection.

PR #270, opened 2026-08-16
add the agent advisor: recoverable-spend and cache-health audit

dinho149:
## Summary

Adds **Agent Advisor** — the fourth agentwatch mode, turning the Agents family's cost meter into a cost advisor. It answers "how much of what my agents cost was avoidable, and why":

- **Read-waste audit** (`agentwatch/waste_audit.py`, vendored/adapted from Headroom's audit tooling, Apache-2.0): re-reads the window's local Claude Code transcripts and sizes identical re-reads, subset re-reads, write read-backs, `cat -n` line-number scaffolding (counted only on reads not already charged to a whole-read class), and stale reads — the last sized as context, never summed into the recoverable headline.
- **Honest pricing**: waste is priced at the window's input-token-weighted blended rate from `pricing.py`; an all-cache window flags `unknown_rate_share=1.0` rather than silently pricing at the fallback tier, and the recoverable headline is capped at the window's measured spend (with a warning) since cached re-reads billed cheaper than fresh input.
- **Cache health** (`agentwatch/cache_signals.py`, adapted from Headroom's detector): structural detection of volatile-shaped content (UUID / ISO-8601 / JWT-shape with a JSON-header check / hex digest) in CLAUDE.md prefix files — counts only, never content, stricter than upstream (no samples) — plus context residency (requestId-deduped turns) and cache-death gaps, rolled into a saturation-resistant 0–100 alignment score.
- **Full surface parity**: engine `run_agent_advisor` (in `advisor.py`, deliberately OFF the Go-mirrored `engine.py` surface), MCP `agents_advisor_run`/`agents_advisor_history`, CLI `yeaboi agents advisor`, TUI Advisor card + page + beta notice + tip, plugin skill `agents-advisor`, parity registry rows.
- **Privacy invariant preserved**: transcripts and prefix files are read locally; only counts, byte totals and file paths reach the artifact, store, exports and screens — enforced by a planted-secret test that scans the artifact AND every DB table.
- Store gains an additive `agent_advisor_reports` table (created idempotently on store open; not mirrored in Go per the dual-maintenance exemption — the sidecar comment in `go/internal/agentwatch/store.go` says so explicitly).
- Docs site: new `docs/agents/advisor.html`, all enumerations updated, SEO regenerated; Apache-2.0 text + upstream NOTICE ship in the wheel under `yeaboi/licenses/`, indexed by `THIRD_PARTY_NOTICES.md`.

An independent fresh-context review ran before this PR; all its should-fix findings are addressed in the second commit. Two deliberate leftovers: `advisor.py` imports five private helpers from `engine.py` (kept — a neutral `_shared.py` would touch the Go-mirrored file's surface for cosmetic gain; revisit if a rename ever bites), and the vendored-source provenance (headroomlabs-ai/headroom v0.35.0) is worth a human spot-check against upstream.

Closes YEA-99

## Test plan

- `make ship-gate` (lint → format-check → full test incl. integration+contract → security → preflight: bundles/docs/go/parity/evaluators/wheel — 6 jobs, all green)
- New: `tests/unit/test_agentwatch_{advisor,waste_audit,cache_signals}.py` (engine math, mechanism fixtures, blended-rate weighting, cap + fallback-rate regressions, JWT false-positive regression, residency turn-dedup regression, DB-wide planted-secret scan)
- Extended: store round-trip for the new kind, render/export/screen tests, MCP inventory, surface-parity + beta-marker guards
- `make package-check` — license files verified present in the wheel
- `make site-check` — docs SEO/sitemap regenerated and clean


dinho149:
Both findings fixed in 439dcedf.

1. **write-readback never fired on real transcripts** — fixed. The containment check now strips the `cat -n` scaffolding off the Read side before comparing against the raw Write body (`_echoes_write`), and the fixture was renumbered to look like an actual Read result (numbered *and* truncated), so the test now disagrees with the bug instead of with the other fixture. A classified readback also asserts its scaffolding is not double-billed.

2. **full Read bodies retained per transcript** — fixed. A cheap pre-scan (`_partial_read_files`, substring-filtered before any JSON parse) finds the files a partial Read targets; only those files' bodies are retained for subset containment — every other read keeps `(sha256, None)`, and identical-repeat stays hash-based. `audit_files` now catches `Exception` rather than `OSError` (logged, counted in `files_skipped`), so the never-raises contract holds for `MemoryError` too. Regression tests cover the hash-only path, the pre-scan, and the non-OSError skip.

Lint clean, 13,232 scoped tests pass.

PR #271, opened 2026-08-16
planning: sprint targeting, sync fixes, team-style ACs, architecture spike, PRD export

dinho149:
## Summary

Five planning-mode features, one commit each, plus a review-fix commit:

1. **Jira/AzDO sprint-sync fixes** (`c0dd23d3` → rebased): kills the "Sprint -1" bug — the `-1` no-tracker sentinel now falls through to the board's max+1 via a new pure `sync_naming` module (consensus naming prefix, advance-past-closed batches). Board sprints are read fully paginated (`maxResults=False`), the scrum board is preferred over kanban, closed sprints are never targeted, and dates post as ISO-8601 datetimes with non-overlapping inclusive ends.
2. **Small projects can add to an existing sprint**: small-mode intake re-enables the sprint question with the board's real active/future sprints as choices; the choice rides `sprint_target_mode`/`target_sprint_name`/`target_sprint_external_id` into sync paths that assign stories and never create a sprint (closed/vanished target = loud error). MCP `plan_sync` gains `target_sprint`.
3. **Team-style acceptance criteria & ticket details**: `AcceptanceCriterion` gains a free-text shape; `resolve_ac_style` (session > `YEABOI_AC_FORMAT`/`--ac-format` > learned team profile > GWT) drives the story-writer prompt and its JSON schema; every renderer/exporter/sync description branches per criterion; learned section headings and custom DoD lists are honoured in tracker descriptions.
4. **Architecture options + opt-in validation spike**: the analyzer proposes 2–3 options with pros/cons and honest confidence (exactly one when pinned by constraint); the user opts in/out of a validation spike (recommended unless confidence is high); an included spike injects deterministically — `[Spike]` story pinned to Sprint 1 (large) or a first `TaskLabel.SPIKE` task (small). Headless/MCP auto-answer by the confidence rule; `--architecture-spike` / `architecture_spike` override.
5. **PRD export**: new `prd_exporter` builds a full Product Requirements Document (MVP scope, verbatim stories in the team's AC style, architecture, phases from sprints, risks) with one LLM call for prose and an honest deterministic fallback — via `plan_export format="prd"`, `plan_publish content="prd"`, `--output prd`, and the TUI export flow.

## Independent review

A fresh-context `code-reviewer` pass found 1 blocker + 6 should-fixes; all are fixed in `address review findings…`: DoD flags now normalise to the plan's resolved list (parse-site sizing + render-side pad/truncate — a custom DoD no longer drops the whole section from synced tickets), the spike story sizes its flags, the sprint-1 pin recomputes `capacity_points`, `_as_lines` can't raise on all-empty dict entries, AzDO iteration datetimes are coerced tz-aware (failure now logs at WARNING), an existing-sprint external id/path is validated against the active/future filter, and the `project-map` skill + parity registry are current. 25 unit tests were added for the previously-untested new functions plus regression tests for every fix.

Four nits intentionally not addressed:
- roadmap-intake's story round-trip (`ui/mode_select/__init__.py`) still serialises GWT-only — unreachable for free-text ACs today (that surface never produces them); left for the roadmap workstream rather than widening this diff into tui-ux-owned code.
- the TUI architecture block's rgb literals match the surrounding file's existing style; a file-wide theme-token sweep belongs to tui-ux.
- the small-mode fallback's reused `jira_status` log line and second tracker round-trip — cosmetic logging/perf on a failure path already logged at WARNING inside the helper.
- the "Generating PRD…" wait reusing the export success-screen builder — a shared-component naming smell, not a behaviour issue.

## Test plan

- `make ship-gate` (lint, format-check, full unit + integration/contract suites, security, preflight: bundles, parity, docs, wheel) — green
- post-review-fix: full unit lane `13398 passed, 47 skipped`, integration/contract `572 passed`, lint clean
- new unit coverage: `test_sync_naming.py`, `test_architecture_spike.py`, `test_prd_exporter.py`, `test_prd_prompt.py`, plus extensions across the sync, nodes, tools, state, sessions, MCP, and parity suites

## DoD notes

- Notion page + Slack post ride the `pr-merged-close-loop` routine on merge, per convention.

Closes YEA-100


dinho149:
Review addressed in c065f9ca — all three should-fix findings fixed:

1. **PRD write encoding** — `export_prd_markdown` now passes `encoding="utf-8"` like every other exporter, with a comment noting the PRD always carries non-ASCII scope markers so a cp1252 default would raise after the LLM call was paid.
2. **`fetch_board_sprints` auth escalation** — restored `_raise_if_auth_error(e, "jira")` and raised the per-state failure log to WARNING, so a 401/403 escalates instead of silently truncating the board under the two new consumers (`_fetch_sprint_targets`, `derive_board_numbering`). New test pins the escalation.
3. **`_parse_spike_reply` negation** — a negation token co-occurring with an include keyword now resolves to skip (`no spike`, `dont add one`, `without the spike`), while a bare negation (`not sure`) stays ambiguous → recommended default. Tests added for all four shapes.

Two of the three nits are also fixed since they were two-line parity/correctness wins: the Jira mixed-AC renderer now numbers GWT triples only (matching the AzDO twin's `gwt_count`), and a closed sprint can no longer shadow a same-named open one in the reuse map (open state wins on duplicate names, with a regression test). The remaining nit (unescaped learned headings in the AzDO HTML description) is left as-is: it matches the surrounding function's existing handling of `story.persona`/`goal`, and escaping one field while its neighbours stay raw would be false comfort — a file-wide escape pass belongs to a dedicated change.

The same commit also adds a user-requested follow-up on the small-project sprint question (a "Backlog (no sprint)" target and explicit create-new labels on the tracker-degraded fallbacks), described in the commit message.

Issue #170, opened 2026-08-06
[docs][analysis] "team-learning" plugin-skill exemption is stale — team-analysis skill already covers its tools


dinho149:
`tests/unit/test_surface_parity.py` (lines ~171-176) carries a recorded `Exempt("no plugin skill yet — tracked gap")` for `team-learning`'s two MCP tools, `team_profile_get` and `team_compare_plan_to_actuals`. Per `cowork/workstreams/analysis.md`, this kind of "tracked" gap is meant to be revisited with a concrete design rather than left standing indefinitely, and both tools already have operational guidance elsewhere — the exemption's stated reason no longer matches reality.

**Impact** 2 · **Effort** S · **Risk** low

---

**What** — Point `team-learning`'s `skill` entry in the `CAPABILITIES` registry (`tests/unit/test_surface_parity.py`) at `"team-analysis"` instead of carrying an `Exempt`, or, if a second standalone skill file is intentionally not wanted, update the `Exempt` reason to say why.

**Why it matters** — Both tools already have operational guidance in the `team-analysis` skill:
- `team_profile_get` — step 1 of `claude-plugin/yeaboi/skills/team-analysis/SKILL.md` ("Check for an existing profile first. Call `team_profile_get`"), lines 8-10
- `team_compare_plan_to_actuals` — step 4 of the same file ("Close the loop... call `team_compare_plan_to_actuals`"), lines 72-74

The exemption's stated reason ("no plugin skill yet") no longer reflects reality — a skill exists and already documents both tools. Leaving it stale means the recorded gap misleads anyone reading it as current.

**Evidence** — `tests/unit/test_surface_parity.py:171-176` (the `Exempt` entry); `claude-plugin/yeaboi/skills/team-analysis/SKILL.md:8-10` and `:72-74` (the coverage that makes it stale).

**Paths** — `tests/unit/test_surface_parity.py` (this is a shared cross-workstream registry file outside the `analysis` workstream's declared paths per `cowork/workstreams/analysis.md`; whichever workstream/human picks this up should coordinate the edit there).

This proposes because it needs a human decision between the two options above, and because the actual edit falls outside this workstream's own paths.

Approve by adding the `claude-implement` label. Reject by closing this issue.

dinho149:
**Reclassified in place** — `cowork:queued`.

This find is on the auto-lane allowlist (`cowork/house-rules.md`, **The auto lane**), so it is not a
[...]

PR #272, opened 2026-08-16
give every workstream a slack voice, and stop timing out unanswered proposals

dinho149:
## Summary

Two changes to the `cowork/` fleet, both about a message reaching the person it is for.

### 1. The evening post is one message per area

`cowork/README.md` says the thirteen maintenance sweeps post **nothing in the channel, ever** — so twelve workstreams had no Slack voice at all, and the one place their work surfaced was a roll-up grouped by *type*, where a fix in `analysis/` arrived as a `[bug]` line between a go-migration wave and a platform chore.

`scripts/cowork_evening.py` now renders one self-contained message per area that moved, headed by that workstream's permanent glyph, and `cron/shipped-standup.md` posts what it is handed — the fleet's rendered-not-composed rule, which matters more here than usual: a model deciding *how many* messages and *which* is one that can post an area twice or drop one silently.

```
🔬 **Analysis** — Sun 16 Aug · 2 merged → `3.9.0rc14`
🧩 **Integrations** — Sun 16 Aug · 1 merged → `3.9.0rc14`
🩺 **Fleet health** — Sun 16 Aug · 1 routine never ran
```

🩺 is its own message because a routine that never ran has no area, and because a thread reply does not notify — which is the failure mode it exists to catch. A post fires on a **change** (merged, opened, newly stuck), never on a standing state, or a long-lived PR re-announces itself nightly.

The seventeen area glyphs live in a new **The area glyphs** table in `cowork/README.md`, guarded by `check_glyphs()` and `--glyphs`. Two are grandfathered (🧭 agents, 🐹 go-migration); 🔐 stays reserved for the security *disclosure* lane, so security's area glyph is 🦺.

**The sweeps themselves still post nothing.** That is a deviation from the literal ask and it is deliberate: a sweep opens a PR at 06:30 and the merge is hours later, so a sweep posting at its own hour announces intent rather than news, and an area would speak twice a day. It also means `cowork/sweep-procedure.md` — constitution — needed no edit; its *a sweep still posts no channel message* rule survives untouched and is now pinned by a test.

`cowork/README.md`'s budget promise moved from "two to four a day, worst case about six" to "three to six, worst case about eight". N messages instead of one is a real cost, and a promise that quietly became untrue is worse than a bigger number.

### 2. A lapsed proposal is not a rejected one

`cron/digest.md` step 4 used to **close** an unanswered proposal at fourteen days. But both dedupe passes read a closed issue as *a closing is a rejection and a rejection is durable*, and they read **state**, not comments — while the `aged-out` marker meant to tell the two apart lives in a comment whose only reader runs on a human's terminal. So a find nobody had time to look at was destroyed by a timer, permanently, and the fleet recorded the difference in the one place nothing it runs can see.

It strips the label and leaves the issue **open** now. The slot reopens with no arithmetic changing (`open_proposals` filters by label), and because nothing closed, no dedupe pass reads a rejection nobody made.

Separately, a find outside a charter's paths is filed against the **owning** workstream's slots rather than the finder's, resolved by `--owner` rather than by eye. Issue #170 is a `test_surface_parity.py` find — platform's file — labelled `workstream:analysis`; it held one of analysis's two slots for ten days over a file analysis may not open.

## Review findings addressed

An independent `code-reviewer` pass found one blocker and seven should-fixes. All are fixed in the second commit:

- **Blocker — 🩺 named its own caller, every evening.** `shipped-standup` is on the 18:00 agenda and checks in *after* posting, so its name could never be in `--checked-in`. Verified against the real agenda; every existing test stubbed `agenda` with a synthetic entry, which is why none of them saw it. New tests run against the real schedule.
- **The no-show window compared a UTC due-time against a *local* ceiling.** Broke two ways silently: a `DISPLAY_TZ` past ~+6 wraps `now` past midnight so 🩺 never fires, and `_local()`'s `" (+1d)"` suffix string-sorts below every bare `HH:MM`. Both sides are UTC now.
- **Red CI came back as a way to be stuck**, and the merged clause carries the check result. `GET /repos/{slug}/commits/{ref}/status` is on the probed allowlist, so this needed no unprobed endpoint. A commit with no statuses claims nothing — the combined API answers `pending` both for a run in flight and for one nothing ever reported on.
- **A `changes requested` crossing fires its own post**, dated by the review's timestamp. Dating it by age left a PR blocked on day 2 unannounced until day 7, or for ever if it merged first.
- **An unreadable release channel is a warning, not `→ no new pre-release`** — that sentence is a claim about the world assembled out of a failure. Three states now, and the title makes no claim when the channel did not answer.
- **`digest.md` step 4 could not run either half of itself.** The 30-day close had nothing that enumerates a lapsed issue (step 1 collects by `cowork:proposal`; a lapse removes exactly that label), so `aged-out` was unreachable. The ⏸️ Held line asked for a lapse date the payload did not carry. Both are Python now — `--lapsed` and `lapses_on` — off step 4's own clock, which is when the question was last *asked*, not when the issue was filed.
- **Install footer gated on `merged`**, `_moment()` coerces a naive `--since`, the paged walk measures GitHub's page length rather than the post-filter one.
- **Coverage gaps closed**: `main()`, `_get()`, `_repo_path()`, `report_owner()` all had no test at all.

Not addressed, deliberately:

- **`declared_paths()` requires backticks** (nit 10) — pre-existing and unchanged by this branch; a `**Paths**` convention change is a charter-wide edit that belongs in its own PR.
- **`/pulls/{n}/reviews` is unpaginated** (nit 13) — 30 reviews on one PR has never happened here, and paging every open PR's reviews nightly costs more than the case is worth.
- **`.days == STUCK_DAYS` is an exact-day equality** (nit 15) — a skipped evening loses that PR's stuck-crossing signal. Real, but the alternative (persisted state between runs) is a bigger change than the fleet's "GitHub issues are the only shared state" rule allows.
- **`cowork_metrics.py` has no reader for the `lapsed:` marker** (nit 16) — worth a follow-up; it is a metrics gap, not a behaviour one.

## Test plan

- `make ship-gate` — lint, format-check, full test suite (unit + integration + contract), security, preflight
- `make test-scoped` after the review fixes and after the rebase — **13388 passed, 47 skipped**
- `make lint`, `make format-check` — clean
- `make cowork-check` — clean
- `uv run pytest tests/unit/test_cowork_evening.py tests/unit/test_cowork_setup.py` — 946 passed
- Each blocker/should-fix has a regression test **verified to fail against the old behaviour**, not just to pass against the new one
- End-to-end render against real GitHub data over the last four days: three messages, correctly separated by area, `ci green` clauses present

Rebased twice: `origin/main` cut `v3.13.0` mid-session, which made `test_release_channel.py` fail on a stale base (pyproject 3.12.1 vs a v3.13.0 tag). No conflicts.

Closes YEA-101


dinho149:
All three fixed — pushed in df53325. Thanks; #1 and #2 were both about halves of this PR's own thesis that I left behind.

**1. Step 4's fourteen-day half now reads the rendered age, and nothing calls `/timeline`.**

You were right that this was the half that did not move. It said *"whose age passes fourteen days"* with no source at all, so a model was ageing seventeen queues by eye — and the `/timeline` recipe under it is refused in a routine session, which makes the failure worse than doing nothing: it falls back to `createdAt` and lapses a bounced proposal up to eight days early, silently, which is exactly what `proposal_clock()` was written to stop one section later.

It reads `--proposal-slots`' `age_days` and `lapses_on` now, both already off the labelling-event clock. I also took the second occurrence you pointed at (⏳ Approved, `digest.md:443`) — `/events` carries the same `labeled` entries and is on the allowlist, so that one was a one-word fix on a call that has been refused every firing. `test_no_routine_tells_anyone_to_call_timeline` keeps it that way, and is scoped to recipes so the paragraph that names the endpoint in order to forbid it still passes.

**2. A lapse is proven now, not inferred from shape.**

This is the one I'm glad you caught — the Python half checked for a recorded `unlabeled` event and the prose half checked for the *absence of two labels*, and I did not notice they were different questions. Your `integration:candidate` example makes it concrete: any issue a human opened and tagged `workstream:<name>` matched, and `cowork:queued` on one of those is a stranger's issue in the unattended build lane with no human verb ever given.

Both prose halves require the same evidence the script does now: `sweep-procedure.md` gates re-labelling on the issue appearing in `--lapsed`, and `cowork-scout.md` returns a `lapsed` field rather than leaving the caller to infer it from the shape. `lapsed: null` re-labels nothing — a duplicate costs less than an unasked build. `TestALapseIsProvenNotInferred` pins both, including one case asserting the code and the prose require the same thing.

**3. The version clause is gated on the rc actually holding one of this area's merges.**

Agreed, and it's the common case rather than an edge one for the reason you gave. I used `pending()`'s `untested_commits` — the commits on `main` that landed after the published tag, i.e. exactly what the rc does not hold — so the test is an exact set membership with no git call and no ancestry inference. `None` (field absent or batch unreadable) claims nothing in either direction.

The title is four states now rather than three: no clause when the channel could not be read, *no new pre-release* when nothing was published, no clause when something was published that this area is not in, and the version when it is. `→ 3.9.0rc14` under a chore was, as you said, a fact inferred in a module whose stated rule is that it never does that.

Verified: `make test-scoped` (4000 passed), `make lint`, `make format-check`, `make cowork-check` all clean, plus an end-to-end render against the last four days of real merges.

PR #274, opened 2026-08-16
add provenance: tamper-evident decision chain, conflict cards, and the audit surface

dinho149:
## Summary

Track B of the repo-scan product roadmap (Track A was #270): every deterministic signal yeaboi surfaces becomes an auditable decision with a verifiable history, and cross-source disagreements become explicit cards instead of silently lowered confidence.

- **`src/yeaboi/provenance/`** — a tamper-evident, hash-chained decision log in the shared `sessions.db`, vendored and substantially adapted from Semantica's `provenance/` package (MIT; attribution in `THIRD_PARTY_NOTICES.md` + `src/yeaboi/licenses/semantica-MIT.txt`). Hardened vs upstream: rows are append-only (no upsert-and-relabel), **every** field is inside the checksum (upstream left `entity_id` and `metadata` editable undetected), composite fields are JSON-encoded in the canonical form so no boundary is malleable, and retraction is a tombstone record — `verify()` detects an edited, deleted, or renumbered row, with ported tamper tests plus the in-place-edit case upstream never covered.
- **Conflicts vocabulary** (`provenance/conflicts.py`) — a fresh implementation imitating Semantica's spec shape with its verified flaws fixed: a real `Severity` enum, value+source paired in one `Claim`, explicit resolution strategies (manual review is the default), credibility-weighted resolution.
- **Standup**: a new engine-layer detector surfaces "the board says Done, but a pull request naming it is still open" as `ConflictCard`s on the report — rendered in the HTML export (new React section, committed bundles), Markdown, and a TUI card. Every standup decision (practice signals with their handles, blocker signals, the confidence verdict with its adjustments, LLM-adjudication drops, conflict cards) is chained; a failed audit write becomes a report warning, never a failed standup.
- **Performance** (the compliance beachhead): 1:1 preps, completions, and six-month reviews each chain one record naming their evidence (ticket keys, the 1:1 history) — counts only, never transcript or note content.
- **`provenance` capability** across surfaces: `run_provenance_audit`/`trace_entity` engines, `provenance_audit`/`provenance_trace` MCP tools, `yeaboi provenance audit|trace` CLI, and a plugin skill; the TUI mode is a reasoned `Exempt` (the cards render inside standup; verify/trace is CLI/MCP-first).

Constraints held: no byte-parity (Go-mirrored) file changes, no `CURRENT_SCHEMA_VERSION` bump (additive idempotent tables), payload rules (severity is a word; the bundle maps tones), all-defaulted frozen artifacts with round-trip tests.

An independent fresh-context review ran pre-PR; all four should-fix findings are fixed in the last commit (honest `llm_mode` on the LLM-free audit tool, count-only prep evidence, unambiguous canonical form with regression tests, structural truncation so `--strict` means what its help says) plus the React-key nit.

## Test plan

- 88 new unit tests: chain tamper scenarios (deleted/edited/renumbered rows, no cascade), checksum coverage/boundary-malleability regressions, conflicts vocabulary + resolution strategies, standup detector + engine wiring (incl. dry-run and failed-write paths), performance recording (incl. transcript-content-never-enters-the-chain), audit engine/renderers, CLI parsing + strict exits, MCP envelope round-trips, store round-trips incl. legacy payloads without the new key.
- `make ship-gate` green: lint, format-check, full suite, security, and all 6 preflight jobs the diff selects (bundles, wheel, evaluators included). Scoped suite re-run after each rebase (13,721 passed).

Closes YEA-103


dinho149:
Both should-fix findings are fixed in e3cf5b00:

1. **Markdown injection through conflict cards** — fixed. `card.title`, `card.detail`, the bare-label branch, and the `source`/`value` strings all route through `_md_label` now, so bracket syntax in an attacker-authored PR title is neutralised before the Markdown reaches Notion/Confluence/Slack; only `_md_link` may mint a link, behind its scheme allowlist. Added `test_markdown_neutralises_link_syntax_in_card_text` with the exact `[Approve access](https://evil.example)` payload from the finding, in a title, an unlinked claim label, and the recommended action.

2. **`verify()` blind to a truncated tail** — fixed with the persisted head anchor you suggested: a `provenance_head` row (max `sequence_id` + its checksum) advances in the same `BEGIN IMMEDIATE` transaction as every append, and `verify()` compares the walk's last record against it, reporting a new `truncated_tail` break reason when the walk falls short — including the dropped-table case, which no longer reads as a fresh install. The store docstring and the plugin skill now also state the honest limit plainly: an adversary who rewrites the rows *and* the anchor together is beyond what a purely local file can prove. New tests: truncated tail (`DELETE … WHERE sequence_id > 3` → break at seq 3, expected 5), dropped table, fresh-empty-chain-stays-valid, and anchor-tracks-invalidations.

On the three nits, taking them as no-action per the review: the `_done_tickets` kind-normalisation asymmetry and the perf `_audit` warn-after-persist ordering are noted for a follow-up sweep, and the unused `resolve`/`ResolutionStrategy` exports are deliberate spec surface for the conflicts vocabulary (the module docstring says why) — if the `dead-code` lens flags them under platform's charter, a `lens-policy.yml` exclusion with that reason is the intended answer.

PR #275, opened 2026-08-16
prior art: batch multi-select carousel replaces the per-repo loop

dinho149:
## Summary

Planning's prior-art repo-relevance step (greenfield intake) asked about each shortlisted repo one at a time — up to 10 graph turns, a card that replaced itself, and every "Not relevant" opening a free-text "why?" turn whose answer permanently blacklisted the repo. It is now one batch:

- **One multi-select carousel in the chat**: a `ChoiceRows(multi=True, carousel=True)` row per candidate (all unchecked), the "You already have this" card re-rendering to preview whichever row is highlighted (↑/↓ and ←/→ while the composer is empty), Space picks, Enter confirms the whole batch in a single graph turn.
- **Kinder rejection semantics**: unchecked = passed over for this project only, nothing written to the suppression ledger. Permanent dismissal is the explicit `X` key (✗ *never suggest*), written as `VERDICT_DOWN` with an empty reason; the free-text reason stage is removed (legacy `"reason"` sessions are read-tolerated and re-asked the batch).
- **One text grammar on every surface** (`_parse_prior_art_answer`): `1 3`, `all`, `none`/`skip`/empty, `!N` bans; ban beats select; any invalid token re-prompts. The node's markdown prompt lists every candidate numbered and documents the grammar, so REPL/form/typed/headless answers work without the widget. The widget submits indices, never labels (a repo named `a, b` would shatter on the old label join). Headless answers `none`.
- **Resume fix**: `_rebuild_transcript` now replays the prior-art card mid-loop instead of the node's raw markdown.
- State field names/shapes unchanged — no migration. `_prior_art_preview` (driver-owned presentation cursor) declared in `ScrumState`, popped on submit and size switch.

Independent review findings all addressed: `isdecimal()` guards the parser against Unicode digits that `int()` refuses; the node's invalid-answer re-prompt now surfaces in chat as prose instead of being swallowed by the static card; `scrum-standards.html` updated to the batch flow; hint sacrifice order fixed (dead entry removed, X shed before Space); direct unit tests for `invalidate_artifact`.

## Test plan

- `make ship-gate` (lint → format-check → full test suite → security → preflight incl. golden evaluators + docs site) — green.
- New/updated tests: grammar parser (incl. `²` re-prompt), batch verdicts + ledger semantics (unchecked writes nothing), legacy reason re-ask, prompt shape, carousel keys (←/→ preview sync, X ban toggle, Space unban, comma-name grammar submit), banned-row rendering, hint overflow at 80 cols, resume card rebuild, re-prompt pass-through, `invalidate_artifact`, stray-preview round-trip.
- DoD: web bundles untouched (no `frontend/` changes); Go parity untouched (off the mirrored surfaces); Notion/Slack fire on merge via `pr-merged-close-loop`.

Closes YEA-104

### DIFF
--- a/docs/docs/scrum-standards.html
+++ b/docs/docs/scrum-standards.html
@@ -165,14 +165,14 @@ gtag('config','G-K9HLPT5ZMP',{client_storage:'none',ads_data_redaction:true});
       </ul>
 
       <h3 id="prior-art">Prior Art (greenfield only)</h3>
-      <p>A greenfield project is rarely built from nothing — the team usually already owns the auth service, the deployment pattern, the design system. When Q2 is answered <strong>greenfield</strong>, the agent inserts one more step between the PTO questions and the intake summary: it ranks the team's own repositories against everything gathered so far and asks about each shortlisted one, with a concrete pitch rather than a bare name.</p>
+      <p>A greenfield project is rarely built from nothing — the team usually already owns the auth service, the deployment pattern, the design system. When Q2 is answered <strong>greenfield</strong>, the agent inserts one more step between the PTO questions and the intake summary: it ranks the team's own repositories against everything gathered so far and offers the whole shortlist at once, each with a concrete pitch rather than a bare name — browse the list and tick the relevant ones in a single answer.</p>
       <ul>
         <li><strong>The repositories come from Team Analysis</strong>, which walks GitHub and Azure DevOps each run and now persists that inventory. Planning reads it offline; a profile captured before this existed says so rather than going quiet.</li>
         <li><strong>Ranking is deterministic</strong> — stack overlap, keyword overlap against Q1/Q3/Q4, recency — and only the top five are enriched with live API calls, so an estate of hundreds costs five reads rather than hundreds.</li>
         <li><strong>The model may only subtract.</strong> It writes the pitch bullets and may drop a candidate it judges irrelevant; it is never given a shape in which to nominate a repository that did not rank.</li>
         <li><strong>Accepted repositories</strong> feed plan generation and appear in the intake summary and every export.</li>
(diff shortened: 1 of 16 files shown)


dinho149:
Worked through the review:

- **Resume hole on the re-prompt path (should-fix): fixed** in 4cb9f374. `prior_art_at` now pins to the newest `AIMessage` whose content is not `_PRIOR_ART_GRAMMAR_HINT`, so a session saved after a rejected typed answer rebuilds with the card replacing the batch prompt and the hint replaying as prose — exactly what the live turn showed. Regression test `test_rebuild_after_a_rejected_answer_cards_the_prompt_not_the_hint` pins the `[AI(prompt), Human, AI(hint)]` shape.
- **`x` claims the first keystroke of free text / ticks lost on re-prompt (nit): answered.** Accepted trade-off, same precedent as Space on multi menus — the composer-empty gate is the one arrow-key rule, the ✗ is visible before Enter, and nothing reaches the ledger. Ticks living on `ChoiceRows` only is also what keeps the widget a pure view of the questionnaire; a re-prompt is rare enough that re-ticking beats a second source of truth.
- **Legacy rejection reasons downgraded to `""` on batch submit (nit): answered.** Deliberate — "the submission is the whole verdict" is the double-write guard, and only sessions serialized mid-loop by an older build can carry a reason. The ledger row's suppression behaviour is identical either way (`is_rejected` keys on `repo_key` alone).
- **`_prior_art_index` written but never read (nit): answered.** Kept for serialization compatibility as noted in `state.py`; the writes stay so a resumed-then-re-serialized session round-trips a stable value. Removing the field is a separate cleanup once no supported session predates the batch flow.

Issue #242, opened 2026-08-12
[blocked] cd-deploy: RemoteTrigger absent from the routine session


dinho149:
`cron/cd-deploy.md` cannot reconcile the fleet. Every firing since 2026-08-10 stops at step 4.

## The cause

The grant is correct — it is the runtime that does not ship the tool.

`RemoteTrigger get trig_01AkW6ojpjKcra8H64R3Astr` returns:

```
"allowed_tools": ["Bash","Glob","Grep","Read","RemoteTrigger","Task","TodoWrite"]
```

which matches `TOOL_OVERRIDES["cd-deploy"]` in `scripts/cowork_setup.py` exactly. The tool is still
absent from the session that runs. **`allowed_tools` filters the toolset; it does not provision a
tool the execution environment does not ship.**

This is the second blocker on that step. The first (`gh` not on PATH) was fixed by the REST
fallback, which now syncs all 31 labels cleanly — fixing it exposed this one. #220 and #236 track
the superseded `gh` problem, not this.

## Impact

- **A merge to `main` does not deploy itself.** The fleet stays current only because a human runs
  `/cowork deploy` locally. `make cowork-check` does not check the account half, so nothing else
  would report the gap.
- **`slack-relay` has the same dependency**, being the only other `RemoteTrigger` holder — Slack
  `pause` / `resume` / `run` are equally dead. It degrades quietly via its own
  "If RemoteTrigger is unavailable" branch instead of alerting, which is why only cd-deploy is
  visible.

## Current state (verified 2026-08-12)

The fleet is **not** drifted, despite the above. Read from a local session, all 23 routines:

```
create: 0   update: 0   ok: 23   orphans: 0
creates_blocked: 0   needs: []   suspicious: false   self_update: null
```

The retired `marketing-weekly` and `integrations-sweep` ids recovered from `cowork/README.md`
history all return 404, so no orphan is hiding past the list's page-1 boundary.

`cowork: cd-deploy` is currently **paused** (`enabled: false`) to stop the alert loop. It needs
resuming once this is fixed.

## Fix options

1. **Provision `RemoteTrigger` in the routine execution environment** — outside this repo.
2. **Move the reconcile into GitHub Actions**, which holds `CLAUDE_CODE_OAUTH_TOKEN` and has no
   egress proxy. `scripts/cowork_setup.py` already owns every comparison and request body, so the
   workflow would only need the snapshot. Unverified crux: whether an `sk-ant-oat` token
[...]

PR #288, opened 2026-08-17
batch the fleet's prs behind a human-tested release batch

dinho149:
## Summary

Changes the cowork fleet's release model from **merge-then-promote** to **batch-then-merge**, so nothing the fleet builds reaches `main` — or users — until a human has installed and exercised the assembled batch.

- **Fleet PRs no longer merge on their own.** Sweeps, campaigns and `codeql-triage` stop arming `gh pr merge --auto`; gate-green fleet PRs wait open, each labelled `semver:none` so `auto-version.yml` bumps nothing on their branches (the batch PR carries the one bump for the whole batch).
- **New `scripts/batch_assemble.py`** (`make batch-assemble`): folds every gate-green fleet PR into a `batch/<date>` branch in a throwaway worktree — one squash commit per constituent, `Closes` lines carried into the body, conflicts and version-bumping constituents skipped and named — then opens a draft PR labelled `release:promotion` and builds the wheel to hand-test. `--close` closes constituents after the batch merges; `--dry-run` assembles locally only.
- **`scripts/beta_signoff.py` retargets at the batch PR**: sha-pinned `<!-- tested: -->` markers (maintainer `authorAssociation` filter unchanged), the per-track/completion marker split preserved, and `beta-promote` now verifies, flips the draft ready, prints `gh pr merge --merge`, and **never merges** — the human's merge is the sign-off, and `publish.yml` classifies it human-lane and cuts the official release from exactly the tested tree.
- **The Slack promote-✅ lane is removed**: `release:promote`, `PROMOTE_RE`-as-verb, and `is_promotion` go; a tombstone regex routes any stale promote-shaped reply to `ask` so it can never reach the `claude-implement` lane. `publish.yml` loses its `issues:` trigger and all marker/left-behind plumbing; the lane classification stays as a backstop. The Monday ask becomes a read-only reminder (its grant now holds only `gh pr list`/`gh pr view`).
- Docs rewritten to match (`release-signoff.md`, `house-rules.md`, `sweep-procedure.md`, campaign routines, `ci-and-release` skill), and new standing guards in `tests/unit/test_release_batch.py` (auto-`ALWAYS`-scoped): assembler mechanics against a throwaway repo, the batch-ships-human-lane invariant, no-routine-arms-auto-merge, `semver:none` documented at every fleet `gh pr create`, merge-not-squash.

## After merge (manual, in order)

1. **Allow merge commits** in repo settings (the batch merges with `--merge`).
2. **Disarm auto-merge on the currently open fleet PRs** (`gh pr merge <n> --disable-auto`) — they are armed under the old model and would merge themselves on the next green CI run.
3. `/cowork deploy` so the routines pick up the new instructions.
4. First cycle watched: `make batch-assemble` → hand-test → `beta-sign-*` → `beta-promote` → merge, and confirm `publish.yml` logs `human merge — cutting the official release`.
5. `publish-beta.yml` is left running (harmless; its rc trails each final) and is slated for retirement after one proven batch cycle.

## Also in this PR: two fleet faults the Slack channel was reporting live

Both were visible in `#yeaboi-claude` while this branch was open, both are in the same
routine-and-report layer this PR already touches, and neither needed a new mechanism.

**`cd-deploy` was paying about a pound per push to say it could not deploy.** Its push webhook
cannot be filtered to a branch (`filter.ref` and `filter.branches` are both rejected — see
`tests/fixtures/cowork_webhook_live.json`), so every push to any branch in a repo full of
worktrees woke a full session: two scripts, then the reconcile step, then the discovery that
`RemoteTrigger` is absent from the runtime (#242), then a 🟡 line about a gap already tracked in an
open issue. Five firings landed in three minutes on 2026-08-17. The tool check moves to **step 2**,
before anything is spent, and the run stops there; the labels/variables half it used to duplicate
is applied by `.github/workflows/cowork-repo-setup.yml` on the same merge, on a runner with admin
scope this session does not have.

**And the heartbeat is now one line a day, not one per firing.** `cowork_checkin.py --quiet-repeat`
prints nothing when a check-in with the same `(name, status)` is already in today's ledger — and
records the run either way, so `make cowork-metrics` still counts every firing. It is a
read-then-write gate with no lock, so it thins rather than deduplicates (concurrent firings both
read a ledger neither has written); that limit is written down where the gate is. `check-in.md`
now grants two exemptions instead of one, and `TestQuietRepeat` pins that a suppressed line is
still recorded.

**The agents digest was posting a proxy 403 as its own content.** `agents-standup` passed
`--github-owners`, which has to expand an owner into repositories first — `GET
/users/{owner}/repos` — and the routine proxy refuses every path that is not repository-scoped
("sessions are bound to their configured repositories", 403). The result was the cryptic message
under a headline saying no agent had done anything. It now names `STANDUP_GITHUB_REPO`, which
takes the single-repository inventory path (already guarded by
`test_ai_usage.py::test_legacy_single_repo_shim_still_fires_without_a_scope`) and makes every
subsequent read `GET /repos/{slug}/…`, which is on the allowlist. A scan that reached nothing now
says so **in the headline**, with the provider note still verbatim underneath. The refused call is
added to `scripts/probe_github_access.py` so the next re-derivation records it as a class, rather
than typed into the fixture by hand.

Guards: `TestADeployThatCannotDeployIsCheap` (the tool check precedes the first script; the skip
names who applies the GitHub half instead; `--quiet-repeat` is wired), the two agents-standup scope
tests, and 12 new `TestQuietRepeat` cases.

## Test plan

- `make ship-gate` green: lint, format-check, full suite, security, all 6 preflight jobs (incl. actionlint on the workflow edits).
- New `tests/unit/test_release_batch.py` (19 tests): squash ordering, conflict skip + clean-worktree recovery, version-bump and empty-constituent skips, `Closes` collection, `gate_green` over both check shapes, human-lane guards, `--close` refusal on an unmerged batch, doc guards.
- Rewritten `tests/unit/test_beta_signoff.py` (35 tests) and updated publish/relay/setup/codeql/zz-guard suites; 14,009 unit tests green on the rebased tree.
- `batch_assemble.py --dry-run` against the real repo: found 5 open fleet PRs, correctly judged none gate-green, exited cleanly creating nothing.
- Independent `code-reviewer` pass: no blockers; the one should-fix (stale committed `uv.lock`) and five nits are all addressed in the second commit.

DoD notes: item 7 (web bundles) does not apply — no front-end changes; items 8/9 fire on merge via `pr-merged-close-loop`.

Closes YEA-110


dinho149:
Both should-fixes are fixed in `2ad8aa9c`, and all three nits with them. fixed=2 answered=0

**1 — the draft batch could never earn a review verdict.** Fixed the smaller of the two ways you named: the batch PR now opens **ready for review**, so `claude-review` posts on the first CI run like any other PR and `pr-feedback` has a verdict to find. The head-sha signature invariant is untouched. `beta-promote` still flips a *hand-opened* draft — a draft cannot be merged at all — but now prints what that costs and why `make batch-assemble` avoids it. Guarded by `TestTheShipIsHumanLane::test_the_batch_pr_is_a_shape_claude_review_will_review`, which asserts no `--draft` in the create call **and** that `claude-review.yml` still skips drafts, so the guard fails loudly rather than silently becoming vacuous if that ever changes.

**2 — `--close` would close any merged PR's `- text (#N)` bullets.** Now requires proof of batch-ness before closing anything: the `<!-- batch: … -->` marker `_body` writes, or the `release:promotion` label. Three new cases cover it, including a round-trip pinning the marker the assembler writes against the regex `--close` reads.

**Nits, all three taken.** The throwaway worktree moved out of `.git` to a predictable temp path (it kept a full checkout plus `uv build` output inside the object store, and could not be created at all from one of this repo's own worktrees, where `.git` is a file). The `worktree prune` is hoisted above the `exists()` guard, so it repairs the case its own comment describes. `release_surfaces.py` and its test now point at `batch_view` instead of the deleted `required_tracks`.

Also in this push, unrelated to the review — two fleet faults the channel was reporting live while this PR was open: `cd-deploy` now checks for `RemoteTrigger` before spending anything (it was buying a full session per push to any branch to report a gap already tracked in #242), `cowork_checkin.py --quiet-repeat` leaves one heartbeat a day instead of one per firing, and `agents-standup` names a repository instead of an owner, because owner discovery is refused by the routine proxy and that 403 was being posted as the digest's own content.

PR #289, opened 2026-08-17
harden the ship gate: deny bash, consent on the toplevel, show the patch

dinho149:
## Summary

Follow-up to #284, addressing all four should-fix findings from that PR's post-CI review (advisory on a hand-shipped branch, so it never held the merge — the findings were real all the same). Two of them were safety claims the code did not actually make good on.

**The agent's capability envelope is now mechanical.** The headless `claude` child was launched with `--disallowedTools "Bash(git push:*)" "Bash(gh:*)"` while the docstring, the run prompt and the changelog all said "no shell". It resolves settings we do not control — the user's `~/.claude/settings.json`, and (its cwd is a checkout of the target repo) that repo's `.claude/settings.json` too. A run against *this* repo would have handed the agent `Bash(make test)`, and a command-pattern deny is porous anyway: `git -c … push`, `git remote set-url` and an edit to `.git/config` all sit outside it. The deny is now the whole `Bash` tool, which outranks every allow rule from every source.

**Consent is checked against the git toplevel.** `fs_policy` containment is `is_relative_to`, so granting `~/code/repo/src` never granted `~/code/repo` — but `worktree.prepare` resolves to the toplevel and every write lands there (`git worktree add` into `<toplevel>/.git`, the later push). Both surfaces now resolve the toplevel *before* the consent prompt and run against it, so the prompt names the directory that is actually touched.

**The gate shows the change.** It rendered `git diff --stat` — eight stat lines in the TUI, the last line only in the CLI — and neither surface showed the worktree path, so the only control before a push was answered on a file count. `ShipRun` now carries `diff_text` (captured by `pipeline.diff_text()`, capped with a trailer naming the git command for the rest); the TUI gate has a scrollable tinted patch pane and the CLI prints the patch inline, both with the worktree path. The stored patch is stripped from the MCP listings — capped per run is not capped per response.

**A gate answers only its own run.** Both loops identified "my run" as "one that was not in the store snapshot I took before launching", which stops being true the moment `YEABOI_AI_MAX_CONCURRENT` > 1 or `YEABOI_AI_BUDGET_DISABLE=1`: another terminal's run opening its gate after the snapshot would be offered to this user to approve and push. `run_ship(..., on_run_id=…)` hands the id back as soon as it is minted, and the heuristic is gone.

An independent pre-PR review then found that the new patch pane could crop the gate's own button row at the enforced minimum terminal size (84x40) — and a gate whose buttons are off screen still answers Enter with "Approve". The gate now measures its rows exactly, gives the elastic sections (validation tail, findings, stat) back in priority order to make room, and drops the pane with a pointer to the worktree rather than ever pushing the buttons off. Its scrolling moved onto the shared `ui/shared/_scroll.py` helpers, which also fixed dead Page Up/Down keys (`read_key` emits `pageup`/`pagedown`, not `pgup`/`pgdn`) and an offset that could run past the last line. The same review's remaining points are in here too: the "nothing reaches origin" claim is now scoped to what argv actually delivers (the validation command still runs a shell in the worktree before the gate, and `--strict-mcp-config` closes the MCP push path), both loops fetch their run by id instead of scanning a bounded newest-first page that concurrency could push it off, run ids carry a random suffix so two same-second runs of one story cannot collide, and the full `--stat` is now the gate's alone rather than every status line's.

Two nits from the same review: `costing.locate_transcript` whitelists the session-id shape (it came from the agent's own JSON envelope and went straight into an `rglob` pattern, where a `*` would have matched and priced an unrelated run), and `worktree.remove()` reports `False` rather than `True` when the registry entry names no checkout. The third nit — `get_ship_log_dir()` having no caller — is deliberate: `paths.py`'s `get_*_log_dir` exports are the API the Observability rule requires, and the hygiene-lens policy already records that shape as unadopted rather than dead.

No schema change (`diff_text` rides the existing run JSON), no new capability, and `ship/` remains off every Go-mirrored surface.

## Test plan

- `make ship-gate` green: lint, format-check, the full suite, security scan, and the preflight jobs this diff needs.
- New tests, each verified to fail without its fix: the whole-tool deny (`test_ship_driver.py`), consent landing on the toplevel and a foreign run never being prompted (`test_cli_subcommands.py`), the patch reaching and scrolling in both gates (`test_ship_screens.py`, `test_ship_render.py`), `diff_text` capping with its trailer and surviving the store round-trip (`test_ship_pipeline.py`, `test_ship_store.py`), `on_run_id` firing exactly once on success and on failure and surviving a raising callback (`test_ship_engine.py`), the glob-shaped session id (`test_ship_costing.py`), and the empty-entry removal (`test_ship_worktree.py`).
- Manual: `yeaboi ship run US-001 --dry-run` renders the canned artifact unchanged.

Closes YEA-111

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "yeaboi"
-version = "3.18.0"
+version = "3.18.1"
 description = "yeaboi — an AI Scrum Master for your terminal: sprint planning, standups, retros, planning poker, 1:1s and delivery reports, plus cost and security posture for your AI coding agents."
 readme = "README.md"
 requires-python = ">=3.11"
--- a/src/yeaboi/agent/state.py
+++ b/src/yeaboi/agent/state.py
@@ -1694,6 +1694,7 @@ class ShipRun:
     phases: tuple[ShipPhase, ...] = ()
     validation: ShipValidation = ShipValidation()
     diff_stat: str = ""  # `git diff --stat` summary shown at the gate
+    diff_text: str = ""  # the capped patch itself — the gate approves a diff, not a file count
     cost_usd: float = 0.0
     transcript_findings: tuple[tuple[str, str, str], ...] = ()  # (kind, severity, label)
     transcript_path: str = ""
--- a/src/yeaboi/changelog_data.json
+++ b/src/yeaboi/changelog_data.json
@@ -4,7 +4,7 @@
     {
       "version": "3.18.1",
       "date": "2026-08-17",
-      "summary": "Ship mode hardening and safety improvements prevent agent capability creep, tighten consent checks, and show you the exact changes before approval. The agent's Bash tool is now completely denied (not just git/gh commands), closing permission-bypass risks from other shell features. Consent is checked against the git toplevel where worktree setup actually lands, so approval names the directory that is actually touched. The approval gate now renders a scrollable patch pane showing your complete diff alongside validation results, with measured button positioning that never leaves buttons off-screen on minimum terminals. Gates answer only their own run by capturing the run id immediately, preventing concurrent runs from presenting stale diffs to the wrong user. Session IDs now carry a random suffix preventing same-second collisions, and removal operations report correctly whether they found an entry.",
+      "summary": "Ship mode hardening and safety improvements prevent agent capability creep, tighten consent checks, and show you the exact changes before approval. The agent's Bash tool is now completely denied (not just git/gh commands), closing permission-bypass risks from other shell features. Consent is checked against the git toplevel where worktree setup actually lands, so approval names the directory that is actually touched. The approval gate now renders a scrollable patch pane showing your complete diff alongside validation results, with measured button positioning that never leaves buttons off-screen on minimum terminals. Gates answer only their own run by capturing the run id immediately, preventing concurrent runs from presenting stale diffs to the wrong user. Run IDs now carry a random suffix preventing same-second collisions, and removal operations report correctly whether they found an entry.",
       "highlights": [
         {
           "text": "Agent permissions hardened: Bash tool is completely denied, not just git/gh commands, closing permission bypasses",
@@ -27,7 +27,7 @@
           ]
         },
         {
-          "text": "Gate buttons stay on screen at minimum terminal size (84x40); elastic sections (findings, stats) yield space gracefully",
+          "text": "The approval buttons always stay on screen, however long the diff or failure log; supporting detail yields space first",
           "areas": [
             "planning"
           ]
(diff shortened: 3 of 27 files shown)


dinho149:
All three fixed — and the first two are exactly the pair you said they were, so thank you for insisting they be fixed together.

**1. `_TITLE_ROWS = 6` → the shared `TITLE_ROWS`.** Confirmed: `render_ascii_text` returns two strings, and `_components.py` already exports `TITLE_ROWS = 2`. My constant was a second, wrong copy of a number that existed. Deleted it and imported the real one, so `budget` is no longer four rows short and the gate stops taking the `patch hidden` branch while four-plus rows of patch would have fitted.

**2. Rows could be wider than the content area, and the wrapped rows were what cropped the buttons.** Also confirmed, and your reading of the interaction was right: with `_TITLE_ROWS` at 6 the accidental slack absorbed the wrapping, so a realistic-content repro passed until finding 1 was fixed. Rather than fix the arithmetic I made wrapping impossible by construction — every gate row now goes through one `_one_row()` helper that clips to the true inner width (`_PANEL_SIDE_COLS = 6`: border 2 + `padding=(1, 2)`) and sets `no_wrap=True, overflow="crop"`. `_gate_line` was the worse half of this, since it never clipped at all; it now clips and truncates with an ellipsis, so a real `~/.yeaboi/ship/worktrees/<repo>/<story>-<stamp>-<suffix>` path can no longer wrap at 84 columns.

New test `test_full_width_patch_lines_and_a_real_worktree_path_do_not_crop_the_buttons` uses 118-column patch lines, a 116-column failure tail and an unclipped worktree path at 84/100/120 columns, asserting both that `Approve` renders and that the panel does not exceed its height. I verified it fails against the original row shape (`PAD` prefixed onto a `line[:width - 10]` clip) rather than only passing against the fix.

**3. ANSI escapes reaching the terminal.** Verified against the installed rich before acting — `STRIP_CONTROL_CODES` is `[7, 8, 11, 12, 13]`, and `Text("a\x1b[2Jb").plain` keeps the ESC. Added `safe_console_text()` in `ship/render.py` (strips C0 and DEL, keeps tab and newline since callers split on newlines) and routed the patch, the diff stat, the validation command and the validation tail through it on both the TUI gate and the CLI renderer. Tests on both surfaces, each verified failing without the strip.

The nits are in too, since they were cheap and you were right about all three: `diff_text`'s `finally` now waits again after `proc.kill()` so a killed `git` cannot linger as a zombie, and the odd expression-statement `close()` is a plain `if`; the give-back loop sheds from the **front** of the failure tail and the stat, so the actual error and the `N files changed` total are the rows that survive; and the release-note prose now says "Run IDs", not "Session IDs", with the `84x40` internal detail replaced by what a user would actually notice. That prose is written by `auto-version.yml` from the commit messages, so I pulled its commit and edited on top rather than force-pushing over it.

`make ship-gate` is green on the result.

PR #292, opened 2026-08-17
redraw the standup timeline, and capture azure devops pr reviews

dinho149:
## Summary

Two things, one reported bug and one rejected UI.

**Azure DevOps PR reviews and comments were not being captured.** Four independent defects, stacked:

- **Reviewer votes were never fetched on the standup path.** The vote-emitting code lived only on the `include_repository=True` analysis branch, which a standup never takes. Votes now come from `VoteUpdate` system threads, which carry a genuine `published_date` and are already in the `get_threads` response — dated votes at zero extra API cost. A PR closed inside the window contributes its final votes at its close date as a secondary source; a stale vote on an old *active* PR with no dated event is deliberately never emitted, because it would repeat in every standup forever.
- **Every AzDO comment on a PR shared the PR's URL**, and the engine dedupes evidence URL-first, so a member's whole review conversation collapsed into one row — or was swallowed by their own PR row. GitHub was unaffected (distinct `html_url`s), which is why the bug looked AzDO-specific. Threads now carry `?discussionId=`, and review rows dedupe in their own namespace (`review|{url}`, mirrored in `go/internal/standup/evidence.go`). Narrowed to reviews on purpose: keying *all* kinds broke the intended ticket-update-vs-issue-row collapse.
- **Caps silently dropped activity.** Thread lookups 25 → 100, project listing 100 → 200, repo-scoped listing 25 → 200, and the comment filter now drops only `system`/`codeChange` instead of every non-`text` type. Eligibility sorts active-before-completed, and logs when the cap bites.
- **The capability manifest claimed the inverse of reality** (votes fetched, comments not), so gap reporting blamed the member instead of the integration.

**The standup page opens on an activity timeline.** The first version of this shipped mute — every title, key and time was hover-only, so a screenshot, a print or an image pasted into Slack conveyed nothing; the axis was drawn true-to-scale so ~45% of the track was empty night; and a merged PR looked identical to a drive-by commit. It was redrawn:

- **Captions on the board.** Landmarks print their own label (`#91 merged`, `YEA-100 In Review`); a burst of commits prints its count. Placement is decided from positions alone — nothing is measured — with landmarks placed first so commits can never crowd out a merged PR.
- **Two tiers.** Landmark discs with glyphs, versus thin strokes at their true positions for commits and comments.
- **A compressed axis.** Events claim footprints, overlapping footprints merge into bursts sized by real duration, and dead stretches pinch into notches. Because that axis is not linear, every notch is labelled with the time it stood for and the panel says so.
- **A rail** carrying each member's span, count, per-kind tally, undated fold and blocked flag, and **threads** joining the lanes that touched the same pull request — matched on the URL path, since GitHub keys reviews on the review id and AzDO on the comment id and neither carries the PR number.

The per-category evidence cap went 8 → 30 (with its Go twin) — at 8 a busy day drew as a thin line.

## Test plan

- `make ship-gate` — lint, format-check, full test suite, security, preflight (go, parity, web, package, eval). Green on this tree.
- `make parity` — 36 pass. The corpus previously exercised **neither** mirrored behaviour: no review shared a PR's URL, and no category held more than 8 rows, so a green run said nothing about either. Both are in the corpus now with a self-guard asserting them.
- Frontend: `npm run typecheck` clean, 1093 vitest tests including a new `timelineScale.test.ts` (the geometry invariants, every degenerate input, and one real URL/key shape per source pinned for the thread matching).
- **Looked at, in a browser**, at 1512px and 676px, in light and midnight, across a dense 8-member day, a sparse 2-member day and a single-event day — with a DOM audit confirming zero overlapping captions, exact lane/overlay geometry, and no horizontal page scroll in every combination. Interactivity (legend filter, tooltips, click-to-jump anchors) verified in the built bundle. That pass is how four layout bugs were found; the first version passed 1056 tests while rendering a row of unlabelled circles.

## Review findings

An independent `code-reviewer` pass raised 1 blocker, 7 should-fix and 8 nits. The blocker (a `gofmt` miss in the parity mirror, which `make parity` does not check but `go-lint` does) and every should-fix are fixed. Two nits are deliberately not addressed, and say so here rather than silently:

- **Comments inside a single thread still share a discussion URL** and fold to one row. One row per discussion is the intent; a `#<comment_id>` tail would split them but make a chatty PR dominate a standup.
- **The widened comment-type filter keeps `unknown`**, which is the SDK's default rather than a hypothetical future value. That is deliberate and pinned by a test — the automation detector (markers + burst + config) still screens service-hook noise downstream, which is the mechanism built for it.

## Definition of Done

- Item 5 (surface parity): **not applicable** — no new capability. This is a bug fix in an existing fetcher plus a rendering change to an existing export surface.
- Item 6 (observability): a `logger.warning` was added where a saturated cap silently truncated. The timeline is client-side rendering with no new Python entry point.
- Items 8 (Notion) and 9 (Slack) fire on merge via `pr-merged-close-loop`.

Closes YEA-112


dinho149:
Both findings taken — the should-fix was right and the nit was not deliberate.

**should-fix — evidence cap 30 vs truncation detector 8 → fixed** (`4f6bcea3`). Correct diagnosis, and the failure scenario is exactly right: `counted > len(evidence)` is the *normal* result of commits nesting under a PR, so `>= evidence_cap` was the whole load-bearing half of that test, and at 30 it stopped bearing it.

Took the "both read one constant" option rather than the one-liner. `MEMBER_EVIDENCE_CAP` lives on `agent/state.py` because the two readers cannot import each other — `gap_taxonomy` is the light module, `engine` the heavy one, and `engine` reaches `transcript_review` which reaches `gap_taxonomy` — and both already import `state`. `engine._member_evidence` and `gap_taxonomy.classify` now default to it, and `evidence.go`'s comment names it so the third copy is findable.

Added the truth-lock you asked for, plus one you didn't: the existing truncation tests were themselves why this drifted unnoticed, since every fixture built exactly eight rows and kept passing while the constant moved underneath them. They are cap-relative now, and `test_does_not_fire_on_a_busy_member_below_the_cap` pins your scenario directly — eleven rows from fifteen activities is busy, not truncated.

**nit — the vote guard keys off the parsed vote → fixed too**, in the same commit. Not deliberate; you're right that `_azdo_thread_vote` returning `None` conflates "not a vote thread" with "vote of 0", and a vote *reset* thread therefore fell through to the comment loop carrying the same bookkeeping the guard exists to drop. Now `_azdo_is_vote_thread` keys off `CodeReviewThreadType`, which is the question actually being asked, with a test for the reset case.

`make ship-gate` green on `4f6bcea3` (lint, format-check, full suite, security, preflight — 6 jobs, 0 skipped, including go/parity/web/package/eval/docs).

PR #298, opened 2026-08-17
surface a live, shareable web board over a supervised ship run

dinho149:
## Summary

Adds a **live, shareable web board over a supervised ship run** — a browser view of yeaboi's story→PR pipeline that the host, and (over a Cloudflare tunnel) teammates, can watch. It is a fourth instance of the existing retro/poker live-board pattern pointed at a ship run, not new infrastructure: it reuses `sharing/` (long-poll transport, event hub, access gate, tunnel) and `web/security.BOARD_CSP` unchanged.

**This ships the read-only board.** Guests watch the five-phase checklist, live agent activity, the diff, the validation verdict and cost. The remote **driver tier** (approve/reject/steer/cancel from the browser) is deliberately deferred — the SQLite gate CAS already makes a browser POST able to unblock a run cross-surface, so the tier is additive and lands behind this safe milestone.

What's here:
- **Stream the agent's output** (`ship/driver.py`, `ship/engine.py`) — `claude --output-format stream-json --verbose` with an `on_line` callback threaded driver → engine → run. `_parse_envelope` is hardened to require `type == "result"` in the tail scan, so a timeout/kill no longer returns a non-result event as the envelope. The TUI gets live output too; a plain run (board off) keeps the one-shot json path.
- **Board backend** (`ship/board.py`, `ship/page.py`, `ship/server.py`) — projects run state into a scrubbed JSON snapshot served read-only over a loopback HTTP server. Every sensitive wire field (`diff_text`, `validation.output_tail`/`command`, `warnings`, `gate_comment`, absolute paths) goes through `standup.gap_issues.scrub`; the activity feed is an allowlist (assistant text + tool *names* only, never tool inputs).
- **React dashboard** (`frontend/src/ship/` → committed `web/static/ship.*`) — token-free boot island, self-contained IIFE bundle, `[data-mode="ship"]` accent (dark + light).
- **TUI wiring** (`ui/mode_select/_ship.py`, `ship/live.py`) — starts the server + tunnel behind `YEABOI_SHIP_BOARD`, never blocking the run.
- **Cross-store plan fix** (`ship/plans.py`) — ship read plans only from the SQLite `SessionStore`, so a plan built in the interactive planning chat (which saves to `persistence.py`) was invisible ("no stories found" over a plan that plainly existed). `plans.py` reconciles both stores for the picker and the engine, preferring the project store then a bounded SQLite scan, and never raises. This also fixes a newer empty session shadowing a real completed plan.

## Test plan

- `make ship-gate` — lint → format-check → test → security → preflight (6 preflight jobs run) — **green**.
- New tests: `test_ship_driver.py` (streaming + envelope hardening), `test_ship_board.py`, `test_ship_server.py`, `test_ship_live.py`, `test_ship_plans.py`, `test_ship_web_page.py` (self-contained / token-free / script-safe page + served-server token-absence). Wire guard: `test_web_wire_shapes.py` pins the ship snapshot shape; `test_web_assets.py` includes the ship bundle.
- Independent fresh-context review (deep tier) run on `git diff origin/main...HEAD`: no blockers. Two should-fix items fixed — the logged host link no longer carries `?token=&admin=` (`live.py`), and `ship/page.py`'s three functions now have unit tests. One nit fixed — the project name is threaded from the picker to the board chrome. Remaining nits (first-block-only activity summary) noted as acceptable for a live feed.
- Verified end to end against real data: the picker now finds a chat-built plan and builds a valid prompt from it.

## DoD notes

- **Notion / Slack (items 8–9)** — not done here; the `pr-merged-close-loop` cowork routine fires them on merge.
- **Review feedback (item 10)** — `claude-review.yml` posts after CI succeeds; the step-4 review above is advisory on a hand-shipped branch.

Closes YEA-115


dinho149:
Thanks — both should-fix findings were real. Fixed in `a53a53bc`.

**1. Phases rendered alphabetically, not in pipeline order** — *fixed*. `board.py:state_snapshot` now projects through a canonical `_PHASE_ORDER` (`setup → implement → validate → gate → finalize`) instead of `sorted()` by id; unreported phases stay absent and an unknown id sorts to the end. The React side maps in array order, so no client change was needed. `frontend/src/test/fixtures/ship.json` regenerated to the correct order (it pinned the bug, as you noted), and `TestPhaseOrder` now (a) asserts the projected order and (b) keeps `_PHASE_ORDER` in lockstep with the TUI's `SHIP_PHASES` tuple so neither drifts silently.

**2. Join code never shown to the host** — *fixed*. The progress screen renders a `join code: XXXX-XXXX` line under the share link once the tunnel is up (`_screens_ship.py`), fed `board.display_code` from `_ship.py`. It rides a separate line from the token-free URL, and only appears when a `share_url` exists (loopback-only shows nothing half-formed). Covered by two render tests in `test_ship_screens.py`. The code no longer lives only in the log file.

Also picked up two of the nits while here (both correct): `driver.py` no longer lets JSONL noise reach `error`/`paused_reason` in stream mode, and the `project-map` skill now documents `SHIP_PORT` + `YEABOI_SHIP_BOARD`.

Left the board-lifecycle nit as-is deliberately: the board is run-scoped and torn down in the `finally` when the engine thread ends, so remote watchers don't see the terminal `approved`/PR-link state. That's a real follow-up (let the board linger briefly to project the outcome), but it belongs with the Phase 4 driver tier — filing under `workstream:planning` rather than widening this read-only PR.

PR #307, opened 2026-08-23
make slack two-way: read reactions and replies back as ceremony control, practice verdicts and corrections

dinho149:
## Summary

yeaboi's Slack integration is write-only today, and structurally so: one `urllib` POST of `{"text": …}` to an incoming webhook with the response body discarded. A webhook answers with the literal string `ok` and no message id, so yeaboi cannot identify its own message — which makes reactions on it unreadable *by construction*. Leg 1 (ceremony cadence, #295) doubled the volume of yeaboi output landing in a team's channel, into a channel that could not answer back.

This makes it two-way. With **both** `SLACK_BOT_TOKEN` and `SLACK_CHANNEL_ID` set, ceremony delivery posts via `chat.postMessage`, gets back `{channel, ts}`, and records an **anchor** row; a launchd/cron interval job reads the last 48 hours every N minutes and applies what it finds. With no token, delivery is byte-for-byte what it is today — two-way is opt-in, not a migration.

**The one rule everything else follows:** *the anchor row is the argument list of the function the event will eventually call.* Slack text never identifies a target and never selects an action; identity is looked up, never parsed. That is what lets all four acts exist with **no LLM anywhere in the inbound path** and no `#<number>` in a message body that could be spoofed.

### The four acts

| Act | Trigger | Calls |
|---|---|---|
| **Control** | ⏸ ▶️ 🚫 on the **post** | `set_enabled` / `set_skip_next` |
| **Verdict** | 👍 👎 on a **signal** reply | `apply_verdict(member=, rule=, run_id=)` — every argument from the anchor |
| **Verb** | a reply that is *exactly* `pause`/`resume`/`skip`/`ack` | the same store calls |
| **Correction** | any other threaded reply | `apply_artifact_edits` as an `OP_NOTE` |

A standup post carries every member's signals in one block of plaintext, so a reaction on it cannot say *which* signal it means — and inferring one is exactly the guess `habits.py` exists to refuse. So the post is a **thread**: one reply per votable signal, each with its own anchor holding `(session_id, run_id, member, rule)` — which is precisely `apply_verdict`'s signature.

### What makes it safe

- **A Slack reply can *add* attributed prose. It can never *change* generated prose.** `OP_SET`/`append`/`remove`/`revert`/`field` are unreachable from `src/yeaboi/slack/` — enforced by an AST walk over the package, not by convention.
- **Authorisation is the first real identity this system has had.** Every author in the edit log is self-declared; a Slack member id on a reaction is attributed by Slack's own servers. `SLACK_ALLOWED_MEMBER_IDS` is machine-wide and hand-curated; **empty ⇒ the poll never calls Slack at all**, and **one malformed entry voids the whole list** — a half-filled allowlist is the more dangerous of the two, because it looks configured. The bot's own id is always excluded, or a token with `reactions:write` plus the ack reaction is a self-authorising loop.
- **Slack never touches launchd or crontab.** A store-only pause is enough — `_guard` already returns `skipped_paused` for exactly this drift and the Ceremonies page shows it. That turns the sharpest privilege here into a visible, recorded, reversible database flag.
- **Polling buys away an entire security surface.** No public endpoint, no `X-Slack-Signature` HMAC, no timestamp replay window, no replay cache. yeaboi initiates every connection. Replay protection falls out of a primary key we needed anyway.
- **`tools/slack.py` exposes no `@tool`, ever**, and `slack poll` is never an MCP tool — an LLM-callable `slack_post_message` would let prompt-injected text in a Jira title reach a team channel.

**Threat statement.** What a malicious or confused *allowlisted* reply CAN do: add one attributed, injection-swept annotation (≤2000 chars) to today's run; pause, resume or skip-next a ceremony in this session's store; cast a thumbs verdict on an anchored signal. What it CANNOT do: change any generated field; reach another session, another artifact kind, or any `(member, rule)` other than the anchor's; delete anything; add or remove an OS job; spend money.

### Also in here

- **A one-shot ceremony skip** — `Ceremony.skip_next` holds an **ISO date, not a bool**, because launchd coalesces missed calendar intervals and a bool skip can be burned by a fire arriving next morning for *yesterday's* slot. Ships standalone value with no Slack at all.
- **`artifact_leases`** closes the YEA-80 race: a 👎 rewrites today's report in place, while an open editable share holds a pre-mutation base copy and would resurrect the signal the vote removed. Lease held ⇒ the durable feedback rows still land, today's row is left to its one writer, and the thread says so. This *mitigates* YEA-80; the full `OP_VERDICT` fix is still its own PR.
- **🔁 re-run is deliberately cut** to leg 3: it is the only act that spends money and the only one whose "did it work" answer arrives minutes later. Cutting it makes every remaining act a free, reversible database write — which is what makes this one PR.

## Where the code decided differently from the plan

1. **The Ceremonies "Slack column" became a page line.** I built the column, rendered it at the enforced 84×40, and it ellipsized `morning-standup` and `Mon–Fri at 09:00` — the two cells the page is most about. Two-way is also machine-wide (one token, one channel, one reader), so the column repeated one word on every row. A fact identical on every row is a page fact. `test_the_row_still_shows_its_name_and_cadence_in_full` keeps the measurement.
2. **The line says nothing on a webhook-only setup.** The first version warned there, and broke `test_no_drift_means_no_warning_line` — the right signal. A webhook is a *supported* configuration, so a permanent `!` telling every fresh install what it lacks is a nag. What it does warn about is the one state with no other symptom: a token configured and no reader installed, every reaction landing in a channel nothing reads.
3. **`apply_inbound_events` has no `scheduled` flag**, though the plan gave it one for symmetry with `run_ceremony`. A poll reads a fixed window and every act is free and idempotent, so its overlap lock and gap notice are unconditional — there is nothing to arm. The CLI keeps `--scheduled` because every installed job's argv already carries it, and now says it is inert.
4. **A real bug the CLI test caught:** `paginate` hands its callback the cursor *positionally* while `users_list` takes it by keyword, so `yeaboi slack members` would have raised `TypeError` on first live use.

## Test plan

- `make ship-gate` — green: `lint` → `format-check` → `test` → `security` → `preflight` (6 jobs run, actionlint skipped — no workflow paths in the diff).
- Rebased onto `origin/main` **first** — the branch was 176 commits behind. Two conflicts: `uv.lock`'s version line (kept upstream's) and the settings sections, a genuine overlap where upstream removed `bedrock` while this branch added `slack` — both intents merged.
- `make parity` untouched and green, as designed: `standup/habits.py` is Go-mirrored but never edited, because the `Excuser` seam already reads the rows `apply_verdict` writes.
- **No `CURRENT_SCHEMA_VERSION` bump** — every new table is additive `CREATE TABLE IF NOT EXISTS`, so Go's ceiling is untouched.
- New: `test_slack_{client,store,grammar,anchors,identity,apply,poller,threads,engine,cli,ops}.py`, `test_artifact_leases.py`. The invariants carry their own tests — no anchor ⇒ no action, replay produces one row and one write, a crash between claim and act is reported and not retried, `apply_verdict` gets the anchor's member and never the reactor's, a void allowlist posts nothing, and the AST walk that keeps `OP_SET` unreachable.

## Review findings

An independent fresh-context review ran against `git diff origin/main...HEAD`. One blocker and five should-fix, all addressed in `e3acf108`; the full reasoning is in that commit message.

The blocker was invisible to the gate: three copies of `_cell` and five of `_last_run_cell` in the ceremonies screen, byte-identical and ~110 lines dead. Only the last binding was live, so behaviour was correct and ruff's F811 stays quiet on it. The five should-fix were real defects — the mention strip running before the unescape that produces the form it guards against; signal replies inviting votes an empty allowlist guarantees are discarded; `slack check` reporting "on" via `auth.test` for a bot nobody invited; `identity.roster` ignoring the `db_path` seam every other call honours; and the four decline outcomes logging outside `mode_log`.

Three of the nits were bugs rather than polish and are fixed too: `*/60` is out of range for cron's 0–59 minute field, `expired()` raised `TypeError` on a naive `now` and that escapes far enough to fail a whole poll, and `parse` rejected the `@U…` form `is_member_id` accepted.

**Answered rather than changed:**

- **`InboundEvent.member` is never populated**, so the first branch of `_author` is dead and the column is always `''`. Kept: the ledger row is the join back to the Slack-verified identity, and the field is where a resolved name belongs once anything resolves one at claim time. Removing the column now would mean re-adding it in leg 3.
- **`collect_events` takes `channel`/`allowed`/`bot_id` and uses none of them.** True, and the un-scoped read is deliberate — anchors carry their own channel, so iterating them is correct even if `SLACK_CHANNEL_ID` changes under a live thread. The parameters are misleading and should go, but removing them touches the poller's seam on the eve of a release rather than at the start of one.
- **The cron marker for a session-less job has a trailing space.** It round-trips through Vixie cron verbatim today. Worth tightening, but the fix is in the shared marker path that every ceremony job also uses, which is a wider blast radius than this PR should take.
- **`refused_before` counts any refusal, not just cap refusals**, so an earlier validator rejection can silence the cap message once. A real edge, and the cap still holds — only its explanation is skipped.
- **A typed `ack` on a post anchor speaks every time, uncapped.** By design: it is the one refusal that teaches a gesture, and going quiet is what makes people believe the channel is not listening. If it becomes noise in practice it wants a per-thread cap, not silence.

## DoD exemptions

- **Item 7 (web bundles): does not apply** — nothing under `frontend/` is touched.

Closes YEA-118


dinho149:
**Fixed — `34886991`.** One finding, and it was right.

`run_poll`'s `db_path` stopped at `_handle`, which called `apply_event(event)` bare. Threaded it through, exactly as described: `_handle(..., db_path=db_path)` → `apply_event(event, db_path=db_path)`.

Two things worth adding to your account of it, both of which make it slightly worse than "a trap for the next caller":

- The split is not symmetric in the way that matters most. The **claim row lands in the handed database while the act lands in the real one**, so the ledger records "handled" on the side that did nothing. Since replay-freedom over the 48h overlapping window rests entirely on `UNIQUE(event_key)`, a poll run this way doesn't just write to two places — it defeats the one invariant the window's whole design depends on.
- Your prediction about the suite was exact: threading the keyword broke **fourteen** tests in `test_slack_poller.py`, every one of them a stub taking `event` alone. That is the clearest possible evidence the seam was never exercised, so I added `TestOnePollReadsAndWritesOneDatabase` — it asserts the applier is handed the poll's own path, rather than just asserting the applier ran.

I did not touch the two notes flagged as no-action. The `_leased()`-before-`apply_artifact_edits` ordering in `_correction` and the `finally`-close on both `EditableSession` paths are both load-bearing and now have your read on record, which is the useful outcome.

Verified: `make lint` clean, scoped lane 14994 passed. `make parity` untouched — nothing here is Go-mirrored.

PR #308, opened 2026-08-23
harden the tunnel sharing path and add an opt-in verified-users tier

dinho149:
## Summary

Hardens the Cloudflare-tunnel sharing path and adds an opt-in **verified-users tier** (`YEABOI_SHARE_MODE=access`), in three tracks plus a setup-UX pass:

**Track 1 — fix the default quick-tunnel path.** Rate limiting and stream caps key on the real visitor (`CF-Connecting-IP`, trusted only while a tunnel is live) instead of collapsing into one `127.0.0.1` bucket; refused hold slots back off server- and client-side instead of busy-polling; `YEABOI_NO_TUNNEL` is honoured by Share Online; credential comparison is total over non-ASCII input and malformed `Content-Length` answers 400 and drops the connection; the cached `cloudflared` binary's digest is re-checked per launch with `YEABOI_CLOUDFLARED_STRICT=1` to refuse unpinned binaries; and the poker duel's server-side mic can only start after the host arms it locally.

**Track 3 — contain `cloudflared`.** The child gets an allowlisted environment (no API keys; no `TUNNEL_*` env reconfiguration), pinned `--loglevel info` / loopback `--metrics`, its own session, and the tunnel hostname is redacted from all logs. Containment, AST no-filesystem, and host-metadata guards keep the properties machine-checked.

**Track 2 — the Access tier.** A named tunnel with generated ingress (validated `originRequest.httpHostHeader` pin), a hostname claim registry, local JWT verification (RS256 allowlist, `aud`/`iss`/`exp`, JWKS cache, fail closed) in `sharing/identity.py`, server-side identity on all four tunnel surfaces (retro, poker, share, ship: pid/byline/admin come from the verified email — impersonation and the `?admin=` bearer both die), and one factory (`open_tunnel`) so `access` mode can never silently fall back to a public quick tunnel. A configured-but-never-switched-on host is refused with a remedy, not published.

**Setup UX.** Sharing is its own Settings tab; a one-time prompt before the first share offers the tier; the wizard is 4 steps with three answers at most — domain (`boards.<domain>` default), then the Access application via a deep link to the create form, with the team name and AUD auto-detected from the hostname's own sign-in redirect; re-runs resume at the first missing fact. `yeaboi --setup-access` is the headless twin. No-domain failures name the fix (dash.cloudflare.com → Add a site).

## Test plan

- `make ship-gate` green (lint, format-check, full suite, security, preflight ×6) after rebasing onto today's `origin/main`.
- Independent fresh-context review of `git diff origin/main...HEAD`: 1 blocker + 7 should-fix, **all fixed** with pinning tests — the blocker (`Permissions-Policy: microphone=()` broke the duel's browser recording) plus ship-server identity gaps, the dead configured-but-off remedy, keep-alive poisoning on bad `Content-Length`, an unbounded identity memo (now 60 s TTL), `~/.yeaboi/bin` 0700 + honest SECURITY.md wording, and stored-tunnel overwrite on partial re-runs.
- Review nits deliberately not addressed: the `X-Forwarded-For` fallback in `client_key` (unreachable behind `CF-Connecting-IP`, kept for the off-Cloudflare proxy case), `save()` skipping empty values (clearing a value is the Settings row's job), the 1.5 s thread-hold on refused stream slots (deliberate server-side throttle), and the O(n log n) lockout eviction at the 1024 cap.
- New/updated suites: `test_sharing_identity` (JWT matrix incl. algorithm confusion, memo TTL), `test_access_tunnel` (tier factory refusals), `test_access_server_identity` + `test_ship_share_tier` (server-level impersonation/join), `test_tunnel_containment`, `test_access_setup` (90: parsers, discovery, resume), `test_access_setup_wizard` (29 walk tests), settings/tips/scope/parity registries.
- Hand-tested on a real Cloudflare account: sign-in, tunnel reuse (`yeaboi`), DNS route, Access app creation, and team/AUD auto-detection against `boards.yeaboi.ai`.

Closes YEA-119
Closes YEA-71

DoD: Notion/Slack fire on merge via pr-merged-close-loop; review feedback is answered after CI's claude-review posts.

https://claude.ai/code/session_019mivuLnWFZuDNaK7eNTcPq


dinho149:
All three findings fixed in c891c30b:

1. **CSRF barrier restored** — `_authed()` on all four servers now requires **both** the verified identity *and* the unguessable board token under the Access tier; the ambient JWT alone (cookie / edge-injected header) is no longer sufficient. The browser already holds the token from `/api/join` (the code gate remains a second factor in this tier), so no front-end change. Tests: `test_identity_alone_is_not_enough_either` replaces the old identity-only-is-served assertion, and every tier test now carries both factors.

2. **AUD validated at the prompt** — new public `access_setup.valid_aud()` (64 hex, case/whitespace tolerant) applied as `validate=` in the wizard and a retry loop in the CLI, with the docstring recording exactly why: `verify()` never reads the aud, so a mistyped tag used to pass "Verified" and then 403 every teammate. Tests cover truncated pastes and a copied application name.

3. **"Pick another hostname" now does what it says** — `_step_hostname` is a loop: the DNS-collision choice, the hard route failure, and the tunnel list/create failures all re-show the domain prompt (with "Try another hostname"/"Try again" labels) instead of returning "back", which walked off the front of the wizard past the skipped sign-in step and closed it with no message. `test_the_collision_choice_re_prompts_instead_of_closing_the_wizard` pins the two-hostname walk.

`make lint` + the scoped lane (15,301 tests) green on the fixes.

PR #310, opened 2026-08-23
make yeaboi installable regardless of the user's python, and lower the floor to 3.10

dinho149:
## Summary

A tester was asked to try the TUI and could not: their system Python was older than 3.11, so
evaluating yeaboi meant a system-level upgrade first. Two independent fixes, because the report
had two causes.

**The root cause was not the floor.** `README.md` offered `uv tool install yeaboi` and
`pipx install yeaboi` as equals. They are not — `uv tool install` fetches its own CPython;
`pipx`/`pip` use the interpreter they run under and hard-fail. Half the audience picked the
coin-flip that cannot work.

**And the floor was self-imposed.** Nothing in the dependency tree ever required 3.11; the
highest floor anywhere in the resolved graph is `>=3.10`. 3.9 stays out of reach — langchain 1.x
floors at 3.10, so reaching it would mean reverting to the 0.x API, a rewrite rather than a bump.

### 1. The install path — `docs/install.sh`

A POSIX-sh bootstrapper served at `yeaboi.ai/install.sh` (Pages-from-branch, no build step), so
the user's own Python stops mattering. Already the URL `cowork/migration/program.md` reserves for
W19's Go distribution, so that wave swaps one file's body and every link keeps working.

Four properties, each tested rather than assumed:

- **`--python ">=3.10"`, a specifier and not a bare version** — a bare version forces a ~30 MB
  download onto people who already have 3.13, and pins everyone to the oldest supported runtime.
  The string is byte-identical to `requires-python`, and a guard couples them.
- **`export UV_PYTHON_DOWNLOADS=automatic` explicitly** — it is uv's default, but a `uv.toml`
  setting `python-downloads = "never"` collapses the whole premise back onto system Python.
- **Never reads stdin** — under `curl | sh` the script *is* stdin.
- **Native Windows is refused with a WSL pointer, not half-served** — `ui/shared/_input.py`
  imports `termios` at module scope, so an install there could not run. The JSON-LD had been
  claiming Windows; narrowed to macOS/Linux.

Also fixed here: the landing page's "Get started →" led to a page whose only install section was
a git clone; `gen_site_seo.py` hardcoded `"Python 3.11+"` where it can now derive it; and
`detect_upgrade_command()` told `uvx` users to run `uv tool upgrade yeaboi`, which errors with
"not installed".

### 2. The floor — 3.11 → 3.10

Shipped as two commits: a behaviour-preserving sweep verifiable on 3.11+, then the flip.

**Three 3.11-only constructs, not two.** `datetime.UTC` and `enum.StrEnum` were expected.
`enum.IntEnum` was not: 3.11 also routed its `str()`/`format()` to the mixed-in type, so
`StoryPointValue.THREE` rendered as its qualified name on 3.10 — into a formatter table and a
serialized artifact. The syrupy baseline caught it. `int.__str__` is the obvious fix and the
wrong one: `int` defines no `__str__`, so it is `object.__str__` and delegates back to Enum's
`__repr__`, turning `"3"` into `"<StoryPointValue.THREE: 3>"`. `int.__repr__` is correct.

**`StrEnum` needs both dunders, and they fail on different versions** — `str()` is wrong on 3.10
*and* 3.11+; `format()` only from 3.11 on. A shim carrying only `__str__` passes its own tests on
3.10 and corrupts every f-string on the version most people run.

**The `UTC` sweep used a `tokenize` codemod, not `sed`** — `UTC` appears in 13+ docstrings and
comments that `sed` would have rewritten; `tokenize` never emits a NAME token inside a string.
It also does not catch aliased access, so a separate AST pass found four `dt.UTC` sites.

**`fromisoformat` greps clean and breaks at runtime.** 3.11 rewrote that parser. Confirmed on a
real 3.10: all three provider formats raise — Jira's `+0000`, GitHub/Notion's `Z`, Azure DevOps's
seven fractional digits. `src/yeaboi/timeparse.py` normalises and **raises `ValueError` like the
stdlib**, so routing ~50 call sites through it is a pure rename against their existing fallbacks.
Six sites hand-rolled `.replace("Z", "+00:00")`; deleting them is the visible payoff.

### 3. CI

`unit` and `integration` **move to 3.10** rather than gaining a matrix. Their names are required
contexts and `strategy.matrix` renames a check — the outage this repo has already had once, and
which `ci.yml:118` documents. The floor is also the version that can actually break, so true
floor coverage costs zero extra job-minutes. Versions above it go to a new **non-required**
`compat` matrix (3.11–3.14, `fail-fast: false`), wired through `test_scope.py`, the scope job's
outputs, the fail-safe fallback and `preflight.py`, with a real `make test-compat` target so the
local gate can run what CI runs.

The `--python 3.10` pin is load-bearing, not cosmetic: runners ship 3.12/3.13 preinstalled and an
unpinned `uv sync` would build a 3.12 venv, leaving a "3.10 job" that proves nothing. Both jobs
assert the interpreter they actually got.

### Two plan premises that did not survive contact

- **`[voice]` was to be gated to `python_version >= '3.11'`**, on the basis that `ctranslate2`
  shipped cp311+ wheels only. It ships **cp310 and cp39** today, and `onnxruntime` is now a
  universal `py3` wheel — the extra resolves cleanly on 3.10 at current versions
  (`ctranslate2` 4.8.1, `faster-whisper` 1.2.1, both latest). No gate added: it would have put a
  false claim in the metadata.
- **`agent/streaming.py` claimed config propagates via contextvars "(Python 3.11+)".** That
  caveat is about *async* propagation, and there are zero `ainvoke`/`astream` calls in `src/` —
  `stream_chat_turn()` is a sync `graph.stream()` on a worker thread. Prose corrected, no code.

## Test plan

Verified on real interpreters, not by inspection.

| What | Result |
|---|---|
| `make ship-gate` | passed — lint, format-check, test, security, preflight (7 jobs, 0 skipped) |
| Full suite on **3.10** | 15,764 passed, 47 skipped |
| Unit lane on **3.11 / 3.12 / 3.13 / 3.14** | 15,407 passed each (`make test-compat`) |
| `make parity` | 36 passed — the Go twins are unaffected by the import sweep |
| `pip-audit` on **both** lock forks | no known vulnerabilities |
| Built wheel → clean 3.10 venv | `yeaboi --version` and `--list-sessions` run |
| `docs/install.sh` end-to-end | on a box where the only visible Python was macOS's system **3.9.6**, uv fetched CPython 3.14 and `yeaboi --version` ran |
| Counterfactual | bare 3.10 confirmed to reject all three provider timestamp formats and to render both enums by name — the shims are load-bearing, not speculative |

New guards, all in `test_scope.py`'s `ALWAYS` because any file can reintroduce the banned
constructs: `tests/unit/test_compat.py`, `test_timeparse.py`, `test_python_floor.py` (ten
surfaces that name the floor, including the OG PNG that nothing else checks), and
`test_install_script.py`.

### Not done here, and why

- **Item 7 (web bundles)** — no `frontend/` changes; nothing to rebuild.
- **Item 5 (surface parity)** — an installer is not a user-facing capability, so no `CAPABILITIES`
  row or `FeatureTip`. `tests/unit/test_surface_parity.py` is untouched and still passes.
- **The `python:3.9-slim` container check** — Docker was unavailable in the shipping environment.
  The local reproduction above establishes the same property. `scripts/release_surfaces.py` gains
  a `bootstrap` row so the live URL is hand-tested each release: Pages serves the script, CI does
  not, so a broken URL is invisible to every test.
- **`https://yeaboi.ai/install.sh` resolving** — the URL only exists once this merges to `main`.

## Independent review

A fresh-context `code-reviewer` pass ran against `git diff origin/main...HEAD` before this PR
opened. **No blockers.** Six should-fix findings and four nits, all addressed in
`address the independent review` — the three worth naming here:

- **`timeparse` had broken the stdlib's error contract for a non-string.** `value.strip()`
  raises `AttributeError` where `fromisoformat(None)` raised `TypeError`, so every rewritten
  call site whose handler names `TypeError` had silently stopped catching it. No live `None`
  path exists today, but the defence was deliberate and is now restored.
- **"Behaves the same on every supported Python" was not true, and nothing tested the gap.**
  Four forms parsed on 3.11+ and raised on 3.10 — a comma fractional separator, the compact
  `20240115T103000` forms, and ISO week dates. All four normalise now, verified by
  differential-testing a 33-case corpus on both 3.10 and 3.14: identical on every input, 18
  rescued on the floor, and no disagreement with 3.14's stdlib anywhere it succeeded.
- **3.11–3.14 were being tested *less* than before, not more.** With the required jobs pinned to
  the floor and `compat` gated on four paths, a change to an ordinary module would have run on
  3.10 and nowhere else — where on `main` today every required job runs 3.11. The trigger is now
  `src/` and `tests/`.

Also found: the enum guard scanned `StrEnum` but not `IntEnum` (a guard passing while the thing
it guards is broken); `curl … | sh` masked a failed download, because POSIX sh has no `pipefail`
and `set -e` takes the last command's status — a 404 exited 0 and the user was then told uv was
"not on PATH"; and an existing uv predating `--python` specifier support was used as-is (the
boundary was measured, not guessed: 0.4.29 rejects, 0.4.30 accepts).

**One finding left as-is, deliberately:** both `parse_date` and `parse_datetime` `.strip()` their
input where the stdlib does not, and `parse_date` is reached from two Go-mirrored modules whose
twin has no such leniency. It is invisible for every stored value those modules read and
`make parity` stays green — recorded here so it is a reviewed divergence rather than a silent one.

### Claude Review

Two findings, both real, both fixed in `024dd4ce` — see the reply on the PR. The temp path the
uv bootstrapper downloaded to was PID-named in a world-writable directory and then executed
(`mktemp` under `$HOME` now, which also makes the header's "writes only under `$HOME`" claim
true). And two validators kept the raw string after validating it, so `--on 20260818` passed,
never equalled the slot's `2026-08-18`, and — because `"2026-08-18" < "20260818"` — never
auto-cleared either, leaving a skip label that could not be removed. That second one is live on
`main` today rather than introduced here, but this branch touches both lines and its subject is
making date parsing uniform.

Closes YEA-121


dinho149:
Both findings were right and both are fixed in `024dd4ce`. Thanks — the second one in particular is a live bug on `main`, not just a hazard.

**1. Predictable temp path — fixed.** `mktemp "$HOME/.yeaboi-uv-installer.XXXXXX"` replaces the PID-named path. Both attacks you named are closed: `O_EXCL` means a pre-created symlink fails the create rather than being written through, and an unguessable name closes the swap window between `curl` returning and `sh` reading. Putting it under `$HOME` rather than `$TMPDIR` also makes the header's "writes only under `$HOME`" claim literally true, which was the right thing to notice — it was documentation describing something the code did not do. Two tests: one asserts the fetched path is under `$HOME` and carries no PID, one that the script contains no `$$` at all.

**2. Dates stored non-canonically — fixed at both sites.** Reproduced exactly as described, and the ceremonies half is worse than "silently never fires": `"2026-08-18" < "20260818"` is `True`, so the auto-clear at `engine.py:145` never fires either and the phantom `skipping 20260818` label cannot be removed by waiting. `store.py` now stores `parse_date(...).isoformat()`, and `_validate_window_dates` returns the canonicalised pair so the ordering check, the `context.py` filter and the period label all see one spelling.

Your scope note is the reason this landed here rather than as a follow-up issue: it is not introduced by this PR — 3.11's `date.fromisoformat` accepts both spellings, so it is live on `main` today — but this branch touches both lines and its whole subject is making date parsing uniform, so fixing it anywhere else would have been artificial. Tests cover `20260818` and `2026-W34-2` for `skip_next`, and the backwards-range case for the window.

**Nit 1 (old uv still first on PATH) — also fixed**, since it was cheap and the logic was genuinely unsound: after the bootstrap the script re-runs `uv_predates_specifiers` rather than only `command -v uv`, and dies with a message naming the actual problem instead of letting the old binary reject `--python '>=3.10'` opaquely.

**Nit 3 (`parse_datetime` strips too) — noted, PR body updated.** You are right that the body only recorded the `parse_date` half of that divergence.

**Nit 2 (`preflight` now runs five interpreters locally) — left as-is, deliberately.** The alternative is a narrower `compat` trigger, which is what the previous review round correctly called out as leaving 3.11–3.14 tested *less* than before. Paying it locally is the cheaper mistake; `COMPAT_PYTHONS` is overridable for a tight loop.

Verified after a rebase onto `origin/main` (which had moved): `make test` 15,441 unit + 572 slow, `make lint`, `make format-check`, `make parity` 36 — all green.

PR #311, opened 2026-08-23
render the performance artifact instead of listing its plaintext

dinho149:
## Summary

Performance mode gathers evidence about one engineer from every other mode — standup's per-member
summaries, the team profile's contributor stats and AI-adoption practice rates, retro cards, poker
votes, reporting delivery — and then showed all of it as a flat list of plaintext bullets, on both
of its output surfaces.

The cause was one function and one habit. `export.py:_perf_args()` forced every artifact through
`sections: [{title, items: string[]}]`, and `evidence.py` computed real figures — 12 of 14 stories,
4.1d cycle time, 62% of changes carrying tests — then stringified them into prose for the LLM prompt
and **discarded the numbers**. No UI ever saw them. `EngineerActivity`, already computed by the
engine, was thrown away the same way.

This threads the real structured data through, so numbers survive as numbers, evidence survives as
linked rows, and both surfaces render the artifact object instead of re-colouring its plaintext.

**The data spine.** New persisted shapes (`PerfMetric`, `EvidenceGroup`, `PerformanceNote`) and new
fields (`metrics`, `evidence_items`, `section_states`, `activity`) on the prep, completion and review
artifacts. `evidence.py` now computes once and projects twice — `_analysis_lines` writes the prompt
prose, `_analysis_metrics` writes the records the UI draws, both from the same dicts — so the figure
quoted to the model is the figure the page renders. Evidence rides as `ActivityEvidence`, standup's
existing shape, which is what made the row rendering nearly free on both surfaces.

**No schema migration.** Artifacts persist as whole-artifact JSON in a `report_json` column with
defaulted rebuilders, so new fields need no column and no `CURRENT_SCHEMA_VERSION` bump.

**TUI.** `_styled()` — sniff the prefix of a plaintext line, pick one of four colours — is deleted.
A new `performance_detail_rows()` renders the artifact: stat tiles, meters for ratios, hand-aligned
evidence rows, and the coverage strip moved from the foot of a long artifact to under the headline,
where a reader can tell a quiet period from an unscanned one. Three confirmed bugs go with it: long
lines wrapped at render time while the scroll math counted them as one row, so **the tail of a long
artifact was unreachable**; the status banner scrolled out of view; and an empty section rendered
nothing at all rather than saying which kind of empty it was.

**Web export and the Cloudflare share.** `Performance.tsx` goes from 93 lines of `<h2>` + `<ul>` to
stat tiles, grouped evidence rows and a coverage strip. Standup's coverage-dot renderer is extracted
to a shared `Coverage.tsx` — the proof the extraction is faithful is that `Standup.test.tsx` did not
change and `export.standup.json` did not move. The share uses the same builders, so it improves
identically; `Anonymize` remains the reduced variant and renders no evidence rows at all.

**Also:** the coverage and evidence-source vocabularies are codegen'd into TypeScript rather than
hand-written in both languages; markdown exports gain `## Measured` and `## Evidence`; the roster
switches to a compact list for a team that does not fit as ASCII art; and `cli.py` prints through the
same renderer, so there is now one renderer for one artifact across two surfaces.

### Three data rules worth reviewing for

- **A metric with no sample is omitted, never emitted as `0`.** An engineer whose spill rate was
  never measured has not spilled 0%. The coverage strip is what explains the absence; a dash is what
  a reader mistakes for zero.
- **No `tone` on any stat.** Colouring a 62% test rate red is the page taking a position it has no
  standing to take. `Cell.tone` on `Profile` stays the one documented exception to the payload rule.
- **Section ids are slugged in Python only.** Two implementations of one slug rule is how a TOC link
  ends up pointing at nothing.

### Deliberately not built

- **Per-claim citations.** Nothing today maps "talking point #2" to a PR — the LLM writes prose from
  a blob. A citation shape was designed and cut: unvalidated, a hallucinated link in a document that
  decides someone's promotion is exactly the failure the coverage vocabulary exists to prevent, and a
  reader cannot tell it from a real one. Evidence is grounded per *source* instead, which still gets
  a reader from a claim to the PRs in one scroll.
- **A review-over-review trend.** `_history_rows` carries no numeric column, so there is no series
  yet. Now that `metrics` is persisted, it becomes a clean follow-on.

### Accepted risk, stated plainly

Evidence rows put live tracker links to a named colleague's commits and PRs on a page that can be
shared over a Cloudflare tunnel. Mitigated by `safe_url`/`safeUrl`, `rel="noopener noreferrer"` (so
the tracker never learns the tunnel URL), and the anonymized path rendering no rows. **Not**
mitigated: whoever holds the tunnel URL can enumerate internal links. Shipped knowingly; a `linked`
flag on the args builder, off for `performance_document`, is a five-line change if we want it later.

## Test plan

`make ship-gate` — lint, format-check, the full suite, security, and the preflight jobs this diff
needs (front-end bundles, TS typecheck, wire guards, packaging).

Tests added: `tests/unit/test_row_ctx.py` (measurement matches `Console.render_lines`; the packer
returns an exact `visible_h`; the oversized-first-item guard), `tests/unit/test_performance_rows.py`,
`Coverage.test.tsx`, plus new classes across the performance engine, store, screen, page and
render/export suites and 15 cases in `Performance.test.tsx`.

Four assertions pin the confirmed bugs: at `scroll_offset=999` the artifact's last section heading is
present (the tail is reachable); `email_summary` renders as wrapped prose carrying no accent style;
the banner survives a full scroll and is dropped — buttons kept — on a short terminal; and panel
height is exact across a height × width × action-count matrix.

Hand-checked: `make run-dry` through the roster and detail view, `uv run yeaboi perf prep` on the CLI
path, and the generated export opened over `file://` (it must render with no server — `ARTIFACT_CSP`
sets `connect-src 'none'`).

## Independent review

A fresh-context `code-reviewer` pass over `git diff origin/main...HEAD` raised eleven findings. **All
eleven are fixed in this branch**; the six it rated should-fix were each a number or a word that
would have been wrong in front of a reader:

1. **A second coverage row for `code`** on the deep-scan path — duplicate keys in the browser, the
   source named twice in the eyebrow, two chips with different states. One source now keeps one row
   (`_amend_coverage`), and a failed extra scan reads `failed` only when nothing else covered the
   source, `partial` when saved history did.
2. **The poker denominator counted the wrong unit** — tickets voted on over sessions held, so five
   sessions out of twenty rendered as `1 of 20`. `_poker_lines` now returns sessions attended.
3. **A capped list was counted** — the tile said `23 items shipped` and the coverage sentence beside
   it said `8`, because the sentence measured the prompt's truncated copy.
4. **Carried evidence was unbounded.** `get_latest_prep` knows nothing about which meeting a prep was
   for, so a 1:1 run months after the last prep inherited its numbers and printed them as facts about
   today — in the artifact that gets emailed to the engineer. Now bounded to `_CARRY_PREP_DAYS`, and
   the prep's date rides along as `evidence_date` and is stated on the page and in the markdown.
5. **`delivery_state` inferred `not_configured` from a real SMTP failure**, in the field whose whole
   purpose was to state a fact rather than infer one. `send_completion_email` (bool) became
   `deliver_completion_email` returning the state — renamed because a caller testing the old result
   for truth would read `"not_configured"` as success.
6. **The `Coverage.tsx`/`evidence_payload` extraction quietly changed standup**: the shared projection
   recursed through itself, so standup's merge-commit dedupe stopped applying to a PR's nested
   commits. The grammar is now passed in as `dedupe_key` and applies at every level.

The five nits are fixed too: `section_states` is no longer carried onto a record (nothing there could
read its keys); tickets gathered on a machine with no session database now produce their metric and
evidence rows; a failed gap-fill says `failed`, not `partly scanned`; the masked artifact is cached
instead of being rebuilt at ~60 fps; and `is_empty` counts the numbers.

One unrelated flake surfaced in the compat lane and is fixed here: `test_retro_tunnel.py`'s
registration-gate test gave a spawned fake binary a 1s deadline, so on a loaded machine it failed
before emitting the URL the test is about — passing for the wrong reason whenever it passed.

## Definition of Done

- Item 5 (surface parity): no new capability — this changes how three existing capabilities render,
  on every surface at once. No `CAPABILITIES` row or `FeatureTip` is due.
- Items 8 and 9 (Notion, Slack) fire on merge via `pr-merged-close-loop`.
- Item 10 (review feedback) is the next sitting — `claude-review.yml` has not run yet.

Closes YEA-123

https://claude.ai/code/session_01R9Pm1zsWeTzpCmZomXVgaY


dinho149:
Both findings were right and both are fixed in `a945676`.

**`evidence.py` — the deep scan claiming items it never contributed.** Correct, and worse than it looks because it is the *paid* path. The note is no longer written by `_gap_fill` at all: it cannot know what survives the caller's caps, and writing a sentence about rows you have already handed off is how it came to describe nine items nobody could see. `_gap_note` is now built by the caller from `len(code_lines) - had_code`, and distinguishes the three real outcomes — rows added, rows found but swallowed by a cap the saved history had already filled, and nothing found. I took the "compute the note from what actually survived" branch rather than interleaving: when saved history already yields 12 code lines there is genuinely no code gap left to fill, so the gathering behaviour was right and only the sentence was lying.

**Delivery and analysis collapsing "no report" into "nobody looked".** Also right, and the sharper half of the finding is that it was wrong in *both* directions. Both sources now draw the same three-way distinction `_coverage_detail` draws for standup, retro and poker: credited → `covered`; a report that exists and names somebody else → `partial`, with "an attribution gap, not an idle period"; nothing to read → `not_configured`. `_read_analysis` returns the profile count so the caller can tell the last two apart, which it previously could not.

**Nit 1 (`matched` counting member rows) — fixed too.** Two lines, and "13 of 12" is the kind of number that costs a reader their trust in every other number on the page. It now counts once per run.

**Nit 2 (`story.sprint` in `ActivityEvidence.repository`) — looked, leaving it.** `Evidence.tsx` renders `repo` as an unstyled muted segment joined with ` · `, so a ticket row reads `Story · Sprint 12 · Done`, which is correct. The wire key is a misnomer for a ticket, but renaming it moves standup's fixture and the TS type for no change to what anyone sees.

Tests added for each: the note's three outcomes, a delivery report crediting someone else, an absent report, and two matching member rows in one standup.

PR #312, opened 2026-08-24
show what a performance run is doing, and export what is on screen

dinho149:
## Summary

Generating a 1:1 prep or a 6-month review is a multi-minute pipeline — `gather_engineer_evidence` sweeps seven sources (tracker REST calls plus the standup/analysis/retro/poker/delivery stores), then one large LLM call, then a store write and a Markdown+HTML export. The TUI showed none of it: it repainted the same actions screen with a static `"Generating 1:1 prep for X…"` for the whole run. Performance was the only long-running mode in the repo without a progress screen.

**A real loading screen.** The three engine runners and `gather_engineer_evidence` now take `on_progress` and emit lifecycle events on the existing `analysis/progress.py` contract — the same one agentwatch's checklist keys on, not a new event shape and not standup's plain-string style, which can never say a phase *finished*. The page paints the shared loading screen (`_build_standup_progress_screen`, re-branded coral) with a live checklist.

Each source's terminal status is **derived from the `SourceCoverage` row that source just produced**, so the rows a lead watches tick past and the coverage strip on the artifact they end up reading cannot disagree. Every declared phase reaches a terminal state: `no_data` on the no-database path and the empty-transcript exit, `failed` on a broken store, `fallback` when the LLM was unavailable.

**Export moved to the output.** The pre-generation row is create-actions only — `1:1 Prep · 1:1 Complete · 6mo Review · Notes · History`. There is nothing to export before an artifact exists.

**Bug fix.** The detail view's `Export`, `Share Online` and `Anonymize` all re-resolved *the engineer's latest artifact* by a fixed `review > completion > prep` ladder instead of using the artifact on screen. Generate a fresh 1:1 prep for someone who already had a 6-month review and the screen showed the prep while Export wrote — and Share Online **published** — the review. All three now go through `_shown_artifact()`; `_performance_latest_artifact` is gone.

**Deliberately out of scope:** cancel. Esc during a run is still drained. Reporting's `cancel_event` / `ReportCancelledError` is the pattern if it is ever wanted.

## Review findings

The independent review raised one blocker and four should-fixes; all are fixed in `347edd85`.

- **blocker — the checklist hid the running row on a 24-row terminal.** The loading screen tail-slices to fit, which drops the *head* — and on a declared checklist that is exactly where the active row sits while everything below it is still pending. During the slowest phase you saw seven pending dots and nothing spinning, strictly worse than the static line this replaced. Windowing now happens inside `_build_activity_progress_rows`, which is the only place that knows which rows are pending: anchored on the running row, filled forward, clamped so look-ahead can never push the active row out. Verified at heights 16–40.
- **should-fix — hardcoded phase ids.** The checklists spelled their component ids as literals; a rename on either side would have shown a permanently pending row and a total never reached, with a green suite. Now pinned two-way against `evidence.SOURCE_*` and `engine.PHASE_*` — the guard caught its own case immediately (`PHASE_CONTEXT`, below).
- **should-fix — the empty-transcript exit** returned before emitting, declaring four phases that never settled. It now settles them as `no_data`.
- **should-fix — untested branches.** Added tests for `_performance_export` / `_performance_document`'s note and unknown-kind branches (called for real, not patched away), and for every `_email_phase` mapping.
- **nit — an unknown coverage word defaulted to `completed`**, i.e. a green ✓, the one mark this feature must never invent. Now falls to `failed`, with a module-level assertion that every word in `COVERAGE_STATES` has a status.
- **nit — dead air in the review.** `gather_ceremony_context` + `_load_framework` ran with no phase, so nothing spun. Added `PHASE_CONTEXT` (`partial` when the ceremony history is unreadable but the framework loaded).
- **nit — war-note comments.** Trimmed per CLAUDE.md; that prose lives in the commit messages.
- **nit — the dead fallback action list lost "Back".** Left as-is deliberately: the runner always supplies `actions`, and the fallback now matches the real row, which genuinely has no Back button (Esc and back-tab cover it). Adding one back would make the dead default disagree with the live row.

## Also in this PR: the Security scan fix

Not original scope, folded in on request because it blocks this PR and every other one.

Moving the repo into the `yeaboi-ai` **organisation** broke the `Security scan` job: since v3, `gitleaks-action` refuses to run for an org without a paid licence (`missing gitleaks license`), and `Security scan` is a required check. The last green run on `main` (`3597e38b`) predates the transfer. No diff could fix it.

Swapped the action for the **gitleaks CLI** — no licence, and the same scanner at the same version (8.21.2) the pre-commit hook already runs locally — pinned by sha256, because a secret scanner fetched over the network is the last thing to take on trust. Verified in both directions rather than assuming: this repo's full history (1,249 commits) scans clean and exits 0, and a planted AWS key / GitHub PAT / Slack token still reports 3 leaks and exits 1. (First attempt at that check used `AKIAIOSFODNN7EXAMPLE`, which is gitleaks' own allowlisted documentation sample and passed vacuously — worth knowing if anyone re-tests this.)

The alternative is a `GITLEAKS_LICENSE` secret, which keeps the action's PR annotations; that needs an account decision, so it stayed on the table rather than being made here.

## Test plan

- `make ship-gate` — lint, format-check, 15,740 unit tests across 3.11–3.14, 572 integration/contract, security, and 7 preflight jobs. Green.
- `make test-scoped` (7,075) + `make lint` after the review fixes. Green.
- `make parity` — 36 passed. Performance is not in the Go dual-maintenance table; `on_progress` is purely additive Python the sidecar does not serve.
- 41 new tests: engine phase lifecycle and fallback/no-data mappings, the coverage-derivation invariant in `evidence.py`, the checklist's pending/settled/windowing rendering at six terminal heights, the two-way phase-id guards, and page-loop tests proving the loading screen paints and that Export/Share act on the artifact on screen.
- Rendered the screen by hand at 16, 18, 20, 23, 30 and 40 rows, at the start, middle and end of a run.

**Not applicable:** DoD item 5 (surface parity) — `on_progress` is in `HIDDEN_ALWAYS`, an injection seam, so MCP and CLI need no change and no registry edit; no new capability is added. DoD item 7 (web bundles) — nothing under `frontend/` changed. The one `uv.lock` line repairs pre-existing drift on `main`, whose `pyproject.toml` already says 3.28.0 while its lock said 3.27.0.

Closes YEA-124


dinho149:
Both nits read, one fixed and one argued.

**Fixed — `_window_phase_rows` anchoring (`_screens_secondary.py`).** You were right, and it reproduced exactly as described. An undeclared id draws after every declared one, so the lowest row on screen is not necessarily the newest: with `deep_scan` on, `gap_scan` is emitted before the model call but rendered last, and at a 16-row terminal the window sat on it —

```
✓ Live scan · 7 item(s)
[7/9] · total 0:09          ← "⠋ Ask the model" off-screen
```

— which is precisely the failure the windowing was added to prevent, re-armed by an id the checklist does not name. `_window_phase_rows` now takes an `anchor_id` chosen by **emission order** (the last `running` event, falling back to the newest event) rather than by position in `component_order`. Verified across heights 16/18/20/23 with a regression test that fails on the old anchor. Worth saying that "unreachable today" was doing less work than it looks: `deep_scan` is already exposed on `yeaboi perf prep --deep-scan` and on the `perf_one_on_one_prep` MCP tool, so the engine emits `gap_scan` today — only the TUI does not ask for it. That is one line away from live.

**Answered — the `uv.lock` / `pyproject.toml` drift.** Not fixing this one, and not because it is cosmetic. It is structurally unfixable from a PR branch: `auto-version.yml` bumps `pyproject.toml` and `changelog_data.json` and deliberately does not touch `uv.lock`, and it re-runs on any follow-up push to the branch. So a commit setting the lock to 3.29.0 would trigger a bump to 3.30.0 and re-open the same one-release gap — a treadmill, not a repair. Nothing enforces it either: no workflow or Makefile target passes `--frozen` or `--locked` (I checked). The line this PR does carry moved the lock 3.27.0 → 3.28.0, which closed a real gap against `main` at the time it was written; the bump then moved the target. If the drift is worth closing it wants a step in `auto-version.yml` that re-locks alongside the bump, which is a `platform` change rather than one this branch should make. Happy to file it if you'd like.

Thanks for checking the windowing arithmetic and the settle-on-every-path walk — the `budget + reserved == max_rows` observation is the invariant I'd have wanted stated, and it is the reason the caller's outer slice is a no-op rather than a second, conflicting trim.

PR #314, opened 2026-08-24
ship any level of the plan, together or one stacked pr per story

dinho149:
## Summary

Ship mode was hardwired to user stories at every layer — `pipeline.find_story` searched `state["stories"]` and nothing else, the TUI picker was a flat story list, the CLI positional was `STORY`, and `plans.latest_plan_with_stories` would not even see a plan without stories. But a plan is three levels deep, and the story is the middle one.

**Ship now takes any node of the plan tree.**

- **Epic, story or task.** A new `ship/scope.py` is the first shared walker of the plan hierarchy: `outline()` builds the picker tree, `find_target()` replaces `find_story`, and links come only from the `feature_id`/`story_id` fields — never from parsing an id, because the task-decomposer prompt still teaches the pre-rename `T-US-E1-001-01` shape while the story writer emits `US-F1-001`, so one real plan can hold both. Orphans land under a synthetic **Ungrouped** epic rather than vanishing from a picker the user can see work in.
- **A task was previously unreachable and is the best target of the three.** `Task.ai_prompt` is already an ARC-structured instruction written for a coding assistant; the prompt now carries it plus the parent story's acceptance criteria, because a task alone says what to change but not why it is correct.
- **Scope, chosen per launch on an epic.** *Together* — one branch, one PR. *Separately* — one **stacked** PR per story, story *k* branching off story *k−1* and targeting that branch. A batch is not a new artifact type: it is N ordinary `ShipRun` rows sharing a `batch_id`, so the gate, the CAS, the worktree registry, the hub and gate-resume all keep working per member unchanged.
- **A failed member stops the batch and keeps what shipped.** Unstarted members are returned but never persisted — writing them would make the hub claim work that never happened. Relaunching the epic **adopts its newest unfinished batch** instead of minting a second one, keyed on the story set the epic has *now*: counting members instead would read a batch holding a rejected attempt as forever unfinished, and a batch whose epic has since grown a story as finished.
- **The launch fuse is the visible constraint, and is surfaced honestly.** `MAX_PER_HOUR=2` / `MAX_CONCURRENT=1` means a five-story epic pauses partway by design; the batch stops with the fuse's own wording on the remaining rows rather than spinning or reporting failure.
- **Ship gets a saved-sessions hub**, so finished runs are reopened, exported and deleted instead of being overwritten by the next run. A batch row opens into its member list, and Export/Delete rebind to the member — bound to the batch row they would have acted on every run in the stack.
- **`resume_ship`** recovers an approval gate stranded by a dead process. `ShipRun.pr_base` is persisted because nothing else can rebuild a stranded member's stack parent: resumed against the default branch, it would re-show every earlier story's diff, which is the whole failure stacking exists to prevent.

**Bug fixed in passing.** `pipeline.build_prompt` read the Given/When/Then triple directly, so every plan generated with `ac_format == "bullets"` sent the coding agent the literal string `"1. Given , when , then ."` — acceptance criteria silently dropped for half of all plans. It uses `AcceptanceCriterion.flat_text` now, with a regression test.

`ShipRun.story_id` becomes `item_id`; compatibility lives entirely at the store boundary (`_dict_to_run` reads both, `listing_dict` re-emits `story_id`), so the `ship_history` MCP payload and the plugin skill's documented contract are unchanged. No schema bump — `ship_runs` stores the artifact as JSON and ship is not Go-mirrored.

Surfaces: CLI gains an `ITEM` positional with `--level` and `--split`; MCP stays read-only and picks the new fields up through `asdict`; the plugin skill and the capability registry are updated.

## Test plan

- `make ship-gate` green: lint, format-check, **15866 passed / 47 skipped**, security, and 7 preflight jobs.
- Compat lane green on 3.11, 3.12, 3.13 and 3.14.
- `test_ship_scope.py` — the tree, orphan collection, and the Ungrouped target being shippable.
- `test_ship_pipeline.py` — prompt fidelity at all three levels, plus the `ac_format == "bullets"` regression.
- `test_ship_engine.py` — member *k*'s `base_ref` and `pr_base` are member *k−1*'s branch; a rejected member stops the batch; a denied budget pauses with the fuse's wording; a relaunch adopts the batch and skips the approved members; and a resumed member still targets its stack parent.
- `test_ship_screens.py` — the tree renders at 84×40, 100×40 and 120×24 with a maximally crowded plan, buttons render on every branch, no row wraps.
- `test_ship_store.py`, `test_ship_hub.py`, `test_ship_page.py`, `test_cli_subcommands.py`, `test_surface_parity.py`.

**Not covered by any unit test, and outstanding:** the hand-run against a real repo — shipping a task, then an epic with Scope=Separately to confirm on GitHub that PR #2 targets PR #1's branch, rejecting member 2 to confirm the batch stops, and relaunching to confirm it resumes at member 2.

The independent pre-ship review's findings are addressed in `c9294be2`.

Closes YEA-122


dinho149:
Both should-fixes were real and are fixed in 2e7d07a. Reproduced each before changing anything.

**1. Per-frame store read in `_open_ship_batch` — fixed.**

Confirmed exactly as described: `members = _members(batch_id)` sat at the top of `while True:`, and the `rows` comprehension below it called `_resumable_reason(m)` per member per frame, which reads the worktree registry and `is_dir()`s the checkout. The sibling code carries the *"Read once per open/reload, never per frame"* comment for this reason and I did not follow it two hundred lines away.

The read is now a `_read()` closure called on open and after the only two keys that can change the list (`c` continue, and returning from a member action). It returns `False` when the batch is empty, which also collapses the old `if not members: return` guard into the same place.

Regression: `test_the_member_list_is_read_once_per_open_not_once_per_frame` counts `ShipStore.batch_runs` calls across eight idle frames and asserts exactly one. I checked it actually catches the regression — moving `_read()` back inside the loop fails it — rather than trusting that it would.

**2. `delete_run` pulling a checkout out from under a live run — fixed.**

The chain is as you traced it: `_await_gate`'s `store.get_run(run.run_id) or run` reads a deleted row as "no resolution yet", and that poll has no deadline by design, so the driving process waits forever on a checkout `worktree.remove` had already `--force`d.

Guarded in the store rather than the hub, so a future delete surface inherits it. `delete_run` now raises `ShipRunBusyError` when the run is at `awaiting_approval` and a *different* live pid owns it. The predicate is your suggested reuse, lifted to `store.driven_elsewhere()` and called from both places — `_resumable_reason` had the only copy, and the point of the guard is that the two agree.

Surfacing it took one line in the shared hub: `_reload(delete_run(run) or "Run deleted.")`. Ship's delete closure returns the reason string; every other mode returns `None` and reads exactly as before. The batch member list carries it back the same way. Four store tests cover it: a live foreign owner refuses and leaves the worktree alone, a dead owner does not block, our own pid never blocks, and a finished run is deletable whoever owns it — only a run at the gate has anyone waiting on it.

**Nits — answered, not changed:**

- **Second run for a stranded member.** Correct, and left alone deliberately. Treating a stranded member as done would make `open_batch` read a batch holding a genuinely failed attempt as finished; treating it as resumable inside `run_ship_batch` would mean the batch sequencer adopting another process's gate, which is the thing fix 2 exists to prevent. Per-member Resume in the hub remains the right route, and a duplicate row is the harmless direction.
- **`_resumable_reason` / `_title` crossing packages.** Agreed on both. `driven_elsewhere` is now the third of these and I promoted that one; the other two stay private until something else wants them.
- **Unescaped fences in `build_ship_markdown`.** Real, and not ship-specific — no exporter in the tree escapes or length-matches its fences, and there is no shared helper to fix it in. Fixing it here alone would make ship the one exporter that differs. Flagging rather than dropping: worth one repo-wide pass under `web-ux`, not a ship-local divergence.

Verified: `make lint` clean, `make test-scoped` 9549 passed, ship suite + surface parity 435 passed.

PR #323, opened 2026-08-26
move the desktop route manifest into contracts/v1/

dinho149:
Phase 4 prep. The desktop extraction needs `routes_manifest.json` reachable as a vendored contract; this moves it before anything is deleted, with `desktop/` still in the tree and green.

Nothing at runtime reads the manifest — `desktop/scripts/gen-routes-manifest.mjs` writes it and the two parity suites read it. Under `src/yeaboi/app/` it would vendor into yeaboi-desktop as `contracts/src/yeaboi/app/routes_manifest.json`; beside `app_http.md` the whole of `contracts/v1/` vendors as one path.

`scripts/test_scope.py`'s `platform` area already claimed `contracts/`, so the file stays mapped; the `desktop` job trigger moves with it.

Verified: `make desktop-check` (typecheck + 186 vitest + the manifest staleness gate) and `make test-scoped` (14,498 passed) both green.


dinho149:
Folded into #324, which carries these two commits as its first phases. CI does not run on a stacked PR here — `ci.yml` filters `pull_request` to `branches: [main]` — and this repo ships multi-phase work as one PR with a commit per phase anyway.

PR #350, opened 2026-08-31
add first-class projects, per-run context toggles, and the solo world

dinho149:
## Summary

Adds **projects** as a first-class scoping primitive, per-run **context toggles**, and reshapes the landing split from two audiences into three worlds.

**Projects.** A projects store (SQLite, schema v31) with session links and a prune guard, and `ProjectScope` — a pull-based resolver that narrows cross-mode reads to one project's sessions. `project_id` threads through the standup, retro, poker, performance and reporting engines and their feeding edges. The surface lands everywhere: engine, MCP tools (`mcp/tools_projects.py`), a `yeaboi project` CLI subcommand, a TUI switcher on the welcome screen's `P` keycap, a `CAPABILITIES` row and a `FeatureTip`.

**Context toggles.** Per-run `context_deps` with a named incognito preset across every surface. Each cross-mode gather is gated behind them — the retro history browser, the poker perspective, the performance evidence gather, and every analysis-profile read. A switched-off source settles its coverage row saying so, so silence never reads as an idle period.

**Three worlds.** The `humans` audience is renamed `team` across config, UI, niko and contracts, and a new `solo` world joins it: a three-way landing split, a solo menu, a flock mascot, and beta badges on solo + agents. Solo runs are tailored — the retro dep is off by default (Solo has no retro mode) and analysis skips the roster.

## Review fixes in this branch

An independent review found one bug the tests could not see, plus five smaller ones; all are fixed in `531e8d21`:

- **`project_id`/`context_deps` never reached a node.** LangGraph drops state keys the schema does not declare, and neither was a `ScrumState` field — so every scoped planning run silently read unscoped, from the TUI and from headless alike. Both are now declared, with a regression test (`test_scope_keys_reach_a_node`) that drives them through a compiled graph rather than asserting on the input dict.
- **The TUI planning session never linked to the active project**, so the planning→standup/reporting edge never fired from the terminal. It now seeds `project_id` and passes it to `create_session`.
- **The perf `--project`/`--context`/`--incognito` flags and the two MCP tool descriptions claimed to be a no-op** while being fully wired into the evidence gather. Corrected — only the engineer-keyed `PerformanceStore` reads stay unscoped.
- **`get_all_history` limited before filtering** on both the retro and standup stores, so a scoped read's window could fill with other projects' runs and hide the project's own. The filter moved into SQL on both.
- `project set-defaults` gained the missing `--context` flag, and now exits 2 instead of printing a success line when given no flags.
- The Solo world's "All on" message no longer claims every source is on (retro stays off), and two prose sites still naming the old lowercase `p` keycap were corrected.

## Notes for the reviewer

- **Rebased onto `origin/main`** (9 commits behind). Three conflicts, all resolved deliberately:
  - Upstream shipped `p` = privacy and `k` = system check while this branch wanted `p` = projects. **Projects was rekeyed to capital `P`** — dispatch, version-row keycap, tip text and both prose sites. `read_key` returns printable characters verbatim and `F` was already an uppercase binding, so `P` is reachable; the version row was checked at 84–200 columns and still crops rightmost-first.
  - `uv.lock` took `origin/main`'s side verbatim — this branch makes no `pyproject.toml` dependency change, so the lock carries no diff.
  - `_components.py` kept upstream's new `PRIVACY_THEME`/`SYSTEM_CHECK_THEME` alongside this branch's rewritten three-worlds comment.
- **`contracts/v1/routes_manifest.json` is a vendored regeneration from yeaboi-desktop**, and it empties the desktop chat command registry (the standalone planning chat folded into the project flow). That is a real parity guard going quiet, so `test_every_terminal_verb_reaches_the_desktop_or_is_exempt` now **skips with its reason** rather than passing on a tautological assert — the run reports the guard as off, and it re-arms the moment the manifest carries one command. Worth a conscious sign-off.
- One reviewer nit is **left unfixed by choice**: in the Solo world `get_context_deps()` always returns a concrete tuple, which overrides a project's `default_context_deps`. That is the point of the solo default (Solo has no retro mode, so a project default enabling retro would be meaningless there), not an accident.

## Test plan

- `make ship-gate` — green: lint, format-check, the full suite (**15,278 passed, 60 skipped**) on Python 3.11/3.12/3.13/3.14, security, and 3 preflight jobs.
- New tests: `test_scope_keys_reach_a_node` (the blocker, driven through a compiled `StateGraph`), scoped `get_all_history` on both stores (including that an empty tuple means no rows, and that the limit counts scoped rows), and three `project set-defaults` CLI cases.

### DIFF
--- a/claude-plugin/yeaboi/skills/niko/SKILL.md
+++ b/claude-plugin/yeaboi/skills/niko/SKILL.md
@@ -37,7 +37,7 @@ produces a new report.
   result to continue one. The thread is persisted locally, so it survives
   restarts and is the same thread the desktop panel and `yeaboi ask` see.
 - **`route`** — where the user is, as a desktop route (`/agents/usage`,
-  `/humans/retro`). It colours the answer toward that screen. Omit it rather
+  `/team/retro`). It colours the answer toward that screen. Omit it rather
   than guessing; Niko says it doesn't know which screen rather than inventing
   one.
 - **`max_rounds`** — tool rounds before Niko must answer. The default of 4 is
--- a/contracts/v1/app_http.md
+++ b/contracts/v1/app_http.md
@@ -11,7 +11,7 @@ contract change — update both sides and this file in the same PR.
 stdout, then nothing else ever:
 
 ```
-YEABOI_APP_READY {"pid":12345,"schema":30,"token":"…","url":"http://127.0.0.1:52341","version":"3.25.0"}
+YEABOI_APP_READY {"pid":12345,"schema":31,"token":"…","url":"http://127.0.0.1:52341","version":"3.25.0"}
 ```
 
 - JSON keys: `url`, `token`, `pid`, `schema` (sessions.py `CURRENT_SCHEMA_VERSION`),
--- a/contracts/v1/app_http.md
+++ b/contracts/v1/app_http.md
@@ -33,7 +33,7 @@ The token never appears in URLs. Missing/wrong token → `401 {"error":"unauthor
 |---|---|---|
 | GET | `/api/health` | unauthenticated; `{ok, pid, version, schema}` |
 | GET | `/api/meta/version` | `{version, schema_version, python, platform}` |
-| GET | `/api/meta/capabilities` | the TUI card inventory verbatim: `{categories, modes, agents, intake}` |
+| GET | `/api/meta/capabilities` | the TUI card inventory verbatim: `{categories, solo, modes, agents, intake}` — `modes` is the Team menu, `solo` the Solo menu |
 | GET | `/api/meta/tips` | `{tips: [{key, text, mode_key, is_new, is_beta, surfaces}]}` |
 | GET | `/api/meta/changelog` | `{entries: [{version, date, headline, summary, highlights[{text, areas, surfaces}]}], areas: [{name, color}]}` |
 | GET | `/api/meta/privacy` | `{headline, statement: [str…], groups: [{key, title}], switches: [{key, env, on_value}], egress: [{key, group, what, where, when, default, off_switch}]}` — the privacy statement and egress-disclosure table, serialized verbatim from `yeaboi.privacy` (the copy owner every surface renders). Carries no capability and is never gated behind one: disclosure must always answer |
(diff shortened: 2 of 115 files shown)


dinho149:
All three fixed in `45028427`, plus an unrelated red the calendar caused overnight.

**1. All-typo `context_deps` → silent incognito.** Confirmed and fixed. `normalize_context_deps` now distinguishes "nothing survived filtering" from "the caller asked for none": a value whose tokens are *all* unknown returns `None` (all on), matching the module's other malformed-input fallbacks and keeping the never-raise contract. Only an explicitly empty value is still incognito. Verified before the fix that `resolve_scope("", context_deps=["retros"])` reported `incognito: True` and `wants("standup"): False`; tests now pin both directions, including at the `resolve_scope` level.

**2. Half-honoured `default_context_deps` in the TUI.** Confirmed and fixed. The seeding moved out of the inline block into `_scope_state_keys()`, which resolves through the public `resolve_scope` — the same precedence the headless path uses — so the project's saved default lands on the `context_deps` state key that `_wants_dep` actually reads. Six tests cover it, including the case you named: a project set to `default_context_deps: []` now seeds `"[]"` rather than leaving the key absent.

**3. Projects page cropping its actions.** Confirmed, and a bit worse than described — at 84×40 *both* actions disappeared from three projects up:

| projects | before | after |
|---|---|---|
| 0–1 | visible | visible |
| 3, 6, 12, 40 | **gone** | visible |

`max_offset` was also 0 at every size, so the rows past the fold were unreachable too; it now reports 13 at 20 projects on a 24-row terminal and 33 at 40. Rather than add a third private copy of the flatten helper, I promoted the ceremonies page's `_render_to_lines` into `ui/shared/_components.py` as `render_to_lines` (per the TUI standard's DRY rule) and pointed both pages at it. The `_screens_secondary.py` copy is a different implementation, so I left it alone rather than fold it in blind — worth a follow-up. Tests now assert the buttons at 1/3/6/12/40 projects and that the page publishes real scroll room.

**Unrelated:** `test_the_last_run_and_the_months_spend_come_back` started failing overnight — it pins a run at `2026-08-17` while `month_spend` defaults to the current month, so it broke when the calendar rolled to September. It's on `main` untouched by this branch and would be red there today too; the date is now computed from the current month.

Not addressed: the `project_id` naming collision in `app_http.md` (a `proj-<8hex>` row id vs. the chat-session id). Real, but it's a doc change to a contract the desktop vendors by sha, so it belongs with that repo's next `contracts-sync` rather than in this branch.

`make ship-gate` green: 15,292 passed, 60 skipped, on 3.11–3.14.

PR #374, opened 2026-09-12
serve the planning room and let every run choose what it reads

dinho149:
## Summary

The engine half of the desktop's redesign (yeaboi-desktop #53): the planning conversation is one stage machine served over HTTP, and every run says which sessions it reads.

- **One stage machine** (`agent/chat_session.py`): review gates, the epic gate, capacity and spike choices and tool-write confirmations are pure, screen-free `reply()`/`advance()` steps with typed events, so the desktop room and the terminal drive one implementation. Every user-authored reply passes the input guardrails.
- **Planning room routes** (`app/routes_chat.py`): the conversation lives in `SessionStore` (schema v35: `sessions_meta.title`, `plan_versions`, `session_labels`; the transcript persisted), with list, advance, update, delete, plan view (`agent/plan_view.py`), accepted-version history and the slash-command list; a plan is named from its description until the analysis names it; the MCP plan tools now see a chat session.
- **Context scope** (`src/yeaboi/context/`): sources, a timeframe (last N sprints, month, quarter, year, custom), a free-text project label and tags, resolved by one precedence rule (caller → the mode's saved config → the last scope used on this machine → unscoped) into the rows a run may read; store filters restored; labels written at every record site into one central `session_labels` table; threaded through every run engine, the MCP tools, the CLI flags, the app run bodies, ceremonies, a TUI context page on every hub and the planning intake, and `GET /api/context/*` routes for the desktop picker.
- `GET /api/sessions/recent` lists thirteen modes with labels; the desktop's regenerated routes manifest is vendored byte-for-byte; parity registries updated (`context` capability, `/planning*`, `/home` as the sessions surface, `/duck` terminal-only, `DESKTOP_CHAT_RETIRED` gone).

Contract: `contracts/v1/app_http.md` "Planning room routes", "Plan view", "Plan versions", "Context scope and labels". Merge yeaboi-desktop #53 first — this PR vendors its manifest.

## Test plan

- `make ship-gate` green on the rebased branch: lint, format-check, unit lane on 3.11–3.14 (17,579 passed), integration + contract (579 passed), security, preflight.
- Two independent review rounds; every blocker and should-fix fixed (epic gate reopening on edit, per-run resolution, delete/update/size waiting for a running turn, guardrails on gate replies, case-insensitive labels, TUI record sites labelling their scope, plan-sync eviction during a turn).
- Live pass of the desktop room against this branch's sidecar (see #53).

https://claude.ai/code/session_0126oQBH3VJHxVFUDGcsJ43b


dinho149:
Both findings fixed in 5312a3c1.

- **should-fix 1 — the context page re-read every store per keystroke.** Typing into Project / Tags / From / To now schedules the preview behind a 0.35 s idle (`TEXT_PREVIEW_DEBOUNCE_S`); leaving the row, Tab, Enter or any non-typing key settles it at once. Three tests in `test_context_page.py` pin it: six keystrokes cost one preview, leaving the row costs one more, the debounce elapsing on a quiet frame costs one more.
- **should-fix 2 — `to_spec()` dropped per-source caps under `all`.** The `all` branch now emits the caps as their own clause (`all@2sprints standup:5,retro:1`), which the parser already read back as `sources=None` + limits; three round-trip tests cover the spec twin, the window, and the dict twin.

The two nits stay as they are: the `sessions_recent` per-store cap under a label filter is the pre-existing 100-row window every hub already lives with (a proper fix is a label-aware store query, filed as a follow-up rather than widened here), and the `delete`/`open` race is closed by the `ChatBusyError` path for the case that can actually re-save.

